PR #8, opened 2020-12-15
change excel import & export field suffix @sys to @System [#176088690]

vixriihi:


projects/laji/src/app/shared-modules/spreadsheet/service/spreadsheet.service.ts
```diff
@@ -268,17 +269,19 @@ export class SpreadsheetService {
...
-    if (!values[GeneratorService.splitCoordinate.system]) {
+    const system = values[GeneratorService.splitCoordinate.system] || values[GeneratorService.splitCoordinate.sys];
+
+    if (!system) {
       return values[GeneratorService.splitCoordinate.N] + ' ' + values[GeneratorService.splitCoordinate.E];
     }
     const suffix = typeof values[GeneratorService.splitCoordinate.N] === 'undefined' ||
       typeof values[GeneratorService.splitCoordinate.E] === 'undefined' ?
-      ' ' + values[GeneratorService.splitCoordinate.system] : '';
+      ' ' + system : '';
```
olzraiti, line 279: The ternary operation is a bit hard to read, I though the string would end up having "system :"

How about:

```
const suffix = (
  typeof values[GeneratorService.splitCoordinate.N] === 'undefined'
  || typeof values[GeneratorService.splitCoordinate.E] === 'undefined'
) ? ' ' + system
  : '';
```

projects/laji/src/app/shared-modules/spreadsheet/service/spreadsheet.service.ts
```diff
@@ -268,17 +269,19 @@ export class SpreadsheetService {
...
 
-    if (values[GeneratorService.splitCoordinate.system] === GeneratorService.splitCoordinateSystem.ykj) {
+    if (system === GeneratorService.splitCoordinateSystem.ykj) {
       return values[GeneratorService.splitCoordinate.N] + ':' + values[GeneratorService.splitCoordinate.E] + suffix;
     }
-    if (values[GeneratorService.splitCoordinate.system] === GeneratorService.splitCoordinateSystem.etrs) {
+    if (system === GeneratorService.splitCoordinateSystem.etrs) {
       return `+${values[GeneratorService.splitCoordinate.N]}+${values[GeneratorService.splitCoordinate.E]}/CRSEPSG:3067`;
     }
     return ('' + values[GeneratorService.splitCoordinate.N]).replace(',', '.') + ',' +
```
olzraiti, line 287: The whole thing is a bit hard to read for at least these reasons:
1. String templates used inconsistently
2. The code is horizontally quite long, making it hard to understand what the string will look like. It would help to grab `splitCoordinate` and `splitCoordinateSystem` to separate variables (so that you could write e.g. `splitCoordinate.N` instead of `GeneratorService.splitCoordinate.N`. You could make it even shorter by grabbing the properties from `splitCoordinate` and `splitCoordinateSystem` to variables, e.g.:
```
const {N, E} = GeneratorService.splitCoordinate;
const {etrs, ykj} = GeneratorService.splitCoordinateSystem;
```

I know refactoring the code wasn't in the scope of this PR, but it's a good idea to clean things on the go :)

olzraiti (changes requested):

projects/laji/src/app/shared-modules/spreadsheet/service/spreadsheet.service.ts
vixriihi: Yes refactoring is fun and it's a good idea to do it as we go :) I refactored the parts the way I think works best atm. I wish there would be more time for refactoring, but currently I do not have that time :(

olzraiti (approved):

PR #9, opened 2020-12-15
Task #175160201

Hvar83:
this is about Quality Filters https://www.pivotaltracker.com/story/show/175160201


wkmor1:
@vixriihi and @Hvar83. I think this is pretty much ready to merge now

vixriihi:
There are few comments in the files that would be good to fine tune still

vixriihi (changes requested):

projects/laji/src/app/+observation/form/observation-form.component.ts
```diff
@@ -423,6 +418,32 @@ export class ObservationFormComponent implements OnInit, OnDestroy {
...
+        this.collectionAndRecordQualityString += element + ':' + event[element].join() + ';';
+      }
+    });
+
+    this.query.collectionAndRecordQuality = this.collectionAndRecordQualityString;
+
+   this.onQueryChange();
+  }
+
+  onSubCategoryChange() {
```
vixriihi, line 444: Seems like this method not used?

projects/laji/src/app/shared-modules/select/metadata-select-with-subcategories/metadata-select-with-subcategories.component.ts
```diff
@@ -0,0 +1,140 @@
...
+export enum SelectStyle {
+  basic,
+  advanced
+}
+
+export interface MetadataSelectPick {
+  [field: string]: string;
+}
+
+export const METADATA_SELECT_VALUE_ACCESSOR: any = {
```
vixriihi, line 30: This is not used

projects/laji/src/app/shared-modules/select/metadata-select-with-subcategories/metadata-select-with-subcategories.component.ts
```diff
@@ -0,0 +1,140 @@
...
+import { MultiLangService } from '../../lang/service/multi-lang.service';
+import { Annotation } from '../../../shared/model/Annotation';
+import { MetadataSelectComponent } from '../metadata-select/metadata-select.component'
+
+export enum SelectStyle {
+  basic,
+  advanced
+}
+
+export interface MetadataSelectPick {
```
vixriihi, line 26: This is not used

projects/laji/src/app/shared-modules/select/metadata-select-with-subcategories/metadata-select-with-subcategories.component.html
```diff
@@ -18,8 +17,6 @@
         (selectedChange)="refreshValue($event)"
       >
       </laji-select-subcategories>
-    </ng-template>
-    <ng-template>
       <select
```
vixriihi, line 20: This select can be removed

projects/laji/src/app/shared-modules/select/metadata-select-with-subcategories/metadata-select-with-subcategories.component.html
```diff
@@ -1,5 +1,4 @@
 <ng-container [ngSwitch]="selectStyle">
```
vixriihi, line 1: this ngSwitch  can be removed now

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/metadata-select-with-subcategories/metadata-select-with-subcategories.component.ts
```diff
@@ -0,0 +1,126 @@
...
+  @Input() filtersName = [];
+  @Input() filtersValues = [];
+
+  @Output() update = new EventEmitter<{id: string[] | string, category: string}>();
+
+  lang: string;
+  active = [];
+  selectedTitle = '';
+  _shouldSort = false;
+  _Options:  [{[key: string]: SelectOptions[]}] | Array<any> = [];
```
vixriihi, line 40: maybe categoryOptions instead.

Typing for is not reflecting the real values most likely should be `{[key: string]: SelectOptions[]}` and then initialised with `{}`. This is used in [here](https://github.com/luomus/laji/pull/9/files#diff-24f96d66f5d15a1249c93013ad5c070d93a6d8d737bd031c20a91aae14b4a03bR111).

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -35,13 +35,17 @@ export class SelectComponent implements OnInit, OnChanges, OnDestroy {
...
   @Input() selected: string[] = [];
+  @Input() selectedGlobalSubCategories: {[key: string]: SelectOptions[]} = {};
   @Input() open = false;
   @Input() disabled = false;
   @Input() outputOnlyId = false;
   @Output() selectedChange = new EventEmitter<string[]|string>();
   @Input() multiple = true;
   @Input() info: string;
   @Input() loading = false;
+  @Input() subCategories = [];
```
vixriihi, line 46: This is used at least in the view to decide the class at least but maybe this could be changed to more generic `@Input() optionClass` and `@Input() optionContainerClass`. Is there really a need for any of the subCategories in this context?

projects/laji/src/app/shared-modules/select/select/select.component.html
```diff
@@ -1,28 +1,38 @@
...
       <ng-container *ngIf="!open">
         <span class="item" *ngFor="let option of selectedOptions; trackBy: track">
-          <label class="lj-item selected" (click)="labelClick($event)">
-            <laji-checkbox [value]="true" (valueChange)="remove(option.id)"></laji-checkbox> {{ option.value}}
+          <label class="lj-item selected" [ngClass]="{'subCategoriesOptions': subCategories.length > 0, '': subCategories.length === 0}" (click)="labelClick($event)">
+            <ng-container *ngIf="subCategories.length === 0 && !isParentFilter else checkboxSubCategoriesGlobal">
+              <laji-checkbox [value]="true" (valueChange)="remove(option.id)"></laji-checkbox> {{ option.value}}
+            </ng-container>
+            <ng-template #checkboxSubCategoriesGlobal>
+              <laji-checkbox [threeState]="isParentFilter" [hasSubCategories]="true" (valueChange)="remove(option.id)" [value]="checkStatus(option.id, (isParentFilter ? 'GLOBAL': subCategory))"></laji-checkbox>{{ option.value }}
```
vixriihi, line 12: having a function as input value is not optimal since it's triggered on each change detection cycle and the checkStatus seems to have quite a lot of logic in it. Could this logic of what is actually selected moved higher up to select-subcategories.component?

vixriihi (changes requested):

projects/laji/src/app/shared-modules/search-filters/three-state-multi-switch/switch-row/switch-row.component.ts
```diff
@@ -17,6 +17,7 @@ export class SwitchRowComponent implements OnChanges {
   @Output() update = new EventEmitter<{id: string, value: any}>();
 
   state;
+  typeCheckbox: number;
```
vixriihi, line 20: instead of number this should be typed to use the CheckboxType Enum (using the enum is more clear)

projects/laji/src/app/shared-modules/select/checkbox/checkbox.component.ts
```diff
@@ -24,17 +31,56 @@ export class CheckboxComponent {
...
     } else {
       this.checkbox.nativeElement.checked = false;
       this.stateClass = 'clear';
     }
   }
 
   changeValue() {
-    this.valueChange.emit(this._value === true ? false : (this._value === false ? (this.threeState ? undefined : true) : true));
+    let valueEmitted;
+    if (this.checkboxType === 2) {
```
vixriihi, line 62: enum values here too would help

projects/laji/src/app/shared-modules/select/select/select.component.html
```diff
@@ -4,26 +4,36 @@
...
           <ng-container [ngTemplateOutlet]="value" [ngTemplateOutletContext]="{option: option}"></ng-container>
         </span>
       </ng-container>
-      <div *ngIf="open" class="lj-item-container">
+      <div *ngIf="open" class="lj-item-container" [ngClass]="{'subCategoriesContainer': checkboxType !== 2, '': checkboxType === 2}">
         <laji-spinner [spinning]="loading"></laji-spinner>
         <div [class.lj-active]="i === selectedIdx" *ngFor="let option of options; let i = index; trackBy: track" style="position: relative;">
-          <label class="lj-item" (click)="labelClick($event)">
-            <laji-checkbox [value]="selectedOptions | includes:option" (valueChange)="toggleValue(option.id)"></laji-checkbox> {{ option.value}}
+          <label class="lj-item" [ngClass]="{'subCategoriesOptions': checkboxType !== 2, '': checkboxType === 2}" (click)="labelClick($event)">
```
vixriihi, line 21: You can remove the checkboxType === 2 all together since adding empty class to class list is not needed
I would also remove all mentions of `subCategoriesOptions` , `subCategoriesContainer` and `subMenu` from this file and use `@Input`s to get the needed classes, that would make it more usable in other places.

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -34,21 +42,23 @@ export class SelectComponent implements OnInit, OnChanges, OnDestroy {
...
   @Input() multiple = true;
   @Input() info: string;
   @Input() loading = false;
+  @Input() checkboxType = CheckboxType.basic;
   @ViewChild('filter') filter: ElementRef;
 
-  selectedOptions: SelectOptions[] = [];
-  unselectedOptions: SelectOptions[] = [];
+  selectedOptions: (SelectedOptions|SelectOptions)[] = [];
+  unselectedOptions: (SelectedOptions|SelectOptions)[] = [];
```
vixriihi, line 57: Why is `SelectedOptions`options here?

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -193,12 +228,24 @@ export class SelectComponent implements OnInit, OnChanges, OnDestroy {
     }
     this.unselectedOptions = [];
     this.options.map(option => {
-      if (selected.includes(option.id)) {
-        this.selectedOptions.push(option);
+      if (this.checkboxType === 2) {
```
vixriihi, line 231: Enum also here would make this more readable

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select/select.component.html
Hvar83: now subcategorydata is selectOptions[] and global is selectedOptions[]. Later i filter that with a different pipe

projects/laji/src/app/shared-modules/select/select/select.component.html
```diff
@@ -4,26 +4,36 @@
     <ng-container *ngIf="!useFilter; else multiWithFilter">
       <ng-container *ngIf="!open">
         <span class="item" *ngFor="let option of selectedOptions; trackBy: track">
-          <label class="lj-item selected" (click)="labelClick($event)">
-            <laji-checkbox [value]="true" (valueChange)="remove(option.id)"></laji-checkbox> {{ option.value}}
+          <label class="lj-item selected" [ngClass]="class.length > 0 ? class[0] : ''" (click)="labelClick($event)">
+            <ng-container *ngIf="checkboxType === 'basic' else checkboxSubCategoriesGlobal">
+              <laji-checkbox [value]="true" (valueChange)="remove(option.id, $event)"></laji-checkbox> {{ option.value}}
+            </ng-container>
+            <ng-template #checkboxSubCategoriesGlobal>
```
vixriihi, line 11: There should not be need to use different checkbox handling since we now have checkboxType as an input to select component

projects/laji/src/app/shared-modules/select/select/select.component.html
```diff
@@ -4,26 +4,36 @@
...
+              <laji-checkbox [checkboxType]="'partial'" [value]="selectedOptions | includesMultiState: option : 'id'" (valueChange)="remove(option.id, $event)"></laji-checkbox>{{ option.id | warehouse | label }}
+            </ng-template>
           </label>
           <ng-container [ngTemplateOutlet]="value" [ngTemplateOutletContext]="{option: option}"></ng-container>
         </span>
       </ng-container>
-      <div *ngIf="open" class="lj-item-container">
+      <div *ngIf="open" class="lj-item-container" [ngClass]="class.length > 0 ? class[1] : ''">
```
vixriihi, line 18: With the classes input suggestion earlier this would become:
```
[ngClass]="classes['optionContainer']"

**Review by vixriihi — CHANGES_REQUESTED** [review]

File: projects/laji/src/app/shared-modules/select/select/select.component.html (hunk 14 above)
Comment (Hvar83) [reply projects/laji/src/app/shared-modules/select/select/select.component.html]: Now not different checkboxType 08e6e3f

File: projects/laji/src/app/shared-modules/select/select/select.component.ts [hunk 16]
```diff
@@ -34,21 +42,24 @@ export class SelectComponent implements OnInit, OnChanges, OnDestroy {
...
   @Input() info: string;
   @Input() loading = false;
+  @Input() checkboxType = CheckboxType.basic;
+  @Input() class: string[] = ['subCategoriesOptions', 'subCategoriesContainer', 'subMenu'];
   @ViewChild('filter') filter: ElementRef;
 
-  selectedOptions: SelectOptions[] = [];
-  unselectedOptions: SelectOptions[] = [];
+  selectedOptions: (SelectedOptions|SelectOptions)[] = [];
+  unselectedOptions: (SelectedOptions|SelectOptions)[] = [];
```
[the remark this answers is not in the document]
Hvar83: Now in select component "SelectedOptions" are not used. I leave still that point open because i need to clean the code and to check more because it seems it's working well, but sometimes i find some bugs. Tomorrow morning i will check more and later i will resolve that point

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -189,17 +238,35 @@ export class SelectComponent implements OnInit, OnChanges, OnDestroy {
...
+        });
       } else {
-        this.unselectedOptions.push(option);
+        if (selected.includes(option.id)) {
+          this.selectedOptions.push({'id': option.id, 'value': option.value, 'info': option.info, 'checkboxValue': true});
+        } else {
+          this.unselectedOptions.push({'id': option.id, 'value': option.value, 'info': option.info, 'checkboxValue': false});
+        }
       }
+
```
vixriihi, line 267: countObj could be replaced with something more descriptive. hmmm I think lines 248 - 268 could be replaced with somethings like this:
```
    this.options.forEach(option => {
      const selectedItem = selected.find(select =>
        (option.id === select) ||
        (option.id === select?.id && (select.checkboxValue === true || select.checkboxValue === false))
      );
      const targetOptions = selectedItem ? this.selectedOptions : this.unselectedOptions;
      const checkboxValue = typeof selectedItem === 'object' && 'checkboxValue' in selectedItem ? selectedItem.checkboxValue : !!selectedItem;

      targetOptions.push({
        ...option,
        checkboxValue
      });
    });

**Review by vixriihi — CHANGES_REQUESTED** [review]

File: projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts [hunk 18]
```diff
@@ -0,0 +1,607 @@
...
+  private arrayEquals(a, b) {
+    return Array.isArray(a) &&
+      Array.isArray(b) &&
+      a.length === b.length &&
+      a.every((val) => b.indexOf(val) > -1);
+  }
+
+  private buildSelectedOptions(filters) {
+    const tmpOptions = this.options;
+    this.activatedRoute.queryParams.subscribe(params => {
```
vixriihi, line 326: There is no unsubscribe to this subscription and this is subscribed on every on change in this component causing this to fire multiple times when toggling any of the components values for example. This is also reading query parameters directly from route instead of using query input. Is there a reason why query input could not be used instead?

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -0,0 +1,607 @@
...
+  @Input() subTitleBase = '';
+  @Input() filtersName = [];
+  @Input() filtersValues = [];
+  @ViewChild('filter') filter: ElementRef;
+
+  @Output() update = new EventEmitter<{id: string[] | string, category: string}>();
+
+  selectedOptions = {};
+  unselectedOptions = {};
+  tmpSelectedOption = {};
```
vixriihi, line 58: could you add types to these three. It would help to clarify initOptions method bellow

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -0,0 +1,607 @@
...
+
+  for (const i in this.selected) {
+    this.selectedOptions[i] = [];
+    this.unselectedOptions[i] = [];
+    const countObj = this.selected[i].filter(item => item.checkboxValue === true).length;
+    this.options[i].map(option => {
+      if (countObj > 0) {
+        selected[i].map(item => {
+          if (item.id === option.id) {
+            this.selectedOptions[i].push({'id': option.id, 'value': option.value, 'info': option.info, 'checkboxValue': item.checkboxValue});
```
vixriihi, line 262: spread operator could be used here also

vixriihi (changes requested):

vixriihi:
sry for only now sending these I just noticed that one of the files diff didn't load for me here and I missed the whole file until now.

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -0,0 +1,525 @@
...
+})
+export class SelectSubcategoriesComponent implements OnInit, OnChanges, OnDestroy {
+  private unsubscribe$ = new Subject<null>();
+
+  @Input() query: WarehouseQueryInterface;
+  @Input() options: [{[key: string]: SelectOptions[]}];
+  @Input() title: string;
+  @Input() filterPlaceHolder = 'Search...';
+  @Input() useFilter = true;
+  @Input() selected: SelectedOptions = {};
```
vixriihi, line 42: There seems to be inconsistencies in the way this property is used. It's good idea to type this and the output (selectedChange) so that we can leverage the type checking

vixriihi (dismissed):

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
vixriihi, line 42: There seems to be inconsistencies in the way this property is used. It's good idea to type this and the output (selectedChange) so that we can leverage the type checking

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -0,0 +1,570 @@
...
+  }
+
+  arrayDiff(a, b) {
+    a = a === undefined ? [] : a;
+    if (b.length >= a.length) {
+      return b.filter(({ id: id }) => !a.some(({ id: itemId }) => itemId === id));
+    } else {
+      return a.filter(({ id: id }) => !b.some(({ id: itemId }) => itemId === id));
+    }
+  }
```
vixriihi, line 567: This method is sometimes returning sometimes a - b and sometimes b - a (depending which is bigger). Better algorithms for this can be found [here](https://stackoverflow.com/questions/1187518/how-to-get-the-difference-between-two-arrays-in-javascript)

vixriihi (changes requested):

projects/laji/src/app/shared/pipe/includes-multi-state.pipe.ts
```diff
@@ -0,0 +1,17 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'includesMultiState'
+})
+export class IncludesMultiStatePipe implements PipeTransform {
+
+  transform(value: any, args: any, property: string, checkValue: string): boolean|undefined {
+    if (value.filter(v => v[property] === args[property]).length > 0) {
```
vixriihi, line 9: if value can be anything calling filter on it without type checking is dangerous. Also the same `value.filter(v => v[property] === args[property]` is ran twice so using internal variable for that could be considered.

This methods signature could be shortened to `transform(value: unknown, option: SelectOptions): boolean|undefined` and target only the specific case of checkbox value.

Then the places where this is used could be changed from `includesMultiState:option:'id':'checkboxValue'` to `checkboxValue:option` (after the name change)

vixriihi (changes requested):

projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
```diff
@@ -34,7 +34,7 @@ <h3 *ngIf="displayTitle">{{ form.title }}</h3>
 <alert *ngIf="editingOldWarning" type="warning">{{ 'haseka.editingOldDocumentWarning' | translate: {date: formatDate(formData.dateCreated)} }}</alert>
 <div class="row" *ngIf="namedPlace" style="margin-top: 10px">
   <div class="col-sm-12">
-    <h3><span *ngFor="let field of namedPlaceHeader; let last = last">{{ field | async }}{{ !last ? ', ' : '' }}</span></h3>
+    <h4><span *ngFor="let field of namedPlaceHeader; let last = last">{{ field | async }}{{ !last ? ', ' : '' }}</span></h4>
```
vixriihi, line 37: This seems to be from some other change set

projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.ts
```diff
@@ -59,7 +59,7 @@ export class DocumentFormHeaderComponent implements OnInit, OnChanges, OnDestroy
 
   ngOnInit() {
     this.updateForm();
-    this.subTrans = this.translate.onLangChange.subscribe(() => {
+    this.subTrans = this.translate.onLangChange.subscribe(res => {
```
vixriihi, line 62: This seems to be from some other change set

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.html
```diff
@@ -0,0 +1,44 @@
...
+  <laji-select
+  [options]="options"
+  [filterPlaceHolder]="'haseka.submissions.filter' | translate"
+  [useFilter]="false"
+  [selected]="selectedOptions['GLOBAL']"
+  [multiple]="true"
+  [info]="info"
+  [open]="open"
+  [disabled]="false"
+  [loading]="options['GLOBAL'] === null"
```
vixriihi, line 13: This is used as and object here, but it's type in the `.ts` specified as `SelectOptions[]`. There is also quite a few errors in the console with text
```
ERROR TypeError: Cannot read property 'GLOBAL' of null

**Review by vixriihi — CHANGES_REQUESTED** [review]

File: projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts [hunk 27]
```diff
@@ -0,0 +1,569 @@
...
+
+  toggle(event, el, typeCheckbox = 0) {
+    if (event.target.classList.contains('no-propagation')) {
+      return;
+    }
+    this.open = !this.open;
+    this.typeCheckbox = typeCheckbox;
+    if (this.open && this.useFilter) {
+      ObservableInterval(10).pipe(takeUntil(this.unsubscribe$), take(1))
```
vixriihi, line 170: timer could be better since you only need this triggered once. Also when is this used I could not find a way to get the part after the first if in here to active?

vixriihi (approved):

PR #10, opened 2020-12-18
Refactor News [#176116558]

vixriihi:
remove the useage of store, introduce facade service to keep state in sync and update the visuals to match other pages


projects/laji/src/app/+news/news.component.ts
```diff
@@ -1,64 +1,70 @@
...
+
+    if (news.featuredImage) {
+      this.headerService.updateFeatureImage(news.featuredImage);
+    }
   }
 
-  ngOnDestroy() {
-    if (this.subTrans) {
-      this.subTrans.unsubscribe();
+  private getNewsDescription(news: News): string {
```
olzraiti, line 51: This is the same algorithm that #12 introduced into header service? Could we use that instead of duplicate code?

olzraiti (changes requested):
Lookin good!

projects/laji/src/app/+news/news.component.ts
vixriihi: Yes as it's states in the #12 commit message :)
> When this is merged #10 can be refined to use HeaderService.getDescription and the whole deprecated store can be removed

PR #20, opened 2020-12-23
Observation search improvement 175160222

MLumme:
Merging collections tree filter into dev.


projects/laji/src/app/shared/service/collection.service.ts
```diff
@@ -61,4 +97,35 @@ export class CollectionService extends AbstractCachedHttpService<ICollectionRang
       switchMap(cols => hasMore ? this.fetchWarehouseCollections(page + 1, cols) : of(cols))
     );
   }
+
+  getCollectionsTree(): Observable<any> {
```
vixriihi, line 101: Would be good to have this typed (since it's a method in a service that's used else where) :)

projects/laji/src/app/shared/service/collection.service.ts
```diff
@@ -61,4 +97,35 @@ export class CollectionService extends AbstractCachedHttpService<ICollectionRang
...
+      tap(data => hasMore = data.lastPage && data.lastPage > page),
+      map(data => data.results || []),
+      map(data => data.map(d => {
+        return {
+          id: IdService.getId(d?.aggregateBy?.['document.collectionId']),
+          count: d.count
+        };
+      })),
+      map(cols => [...collections, ...cols]),
+      switchMap(cols => hasMore ? this.fetchWarehouseCollections(page + 1, cols) : of(cols)));
```
vixriihi, line 129: should this be `this.getCollectionsAggregate` instead of `this.fetchWarehouseCollections`?

projects/laji/src/app/shared/service/collection.service.ts
```diff
@@ -61,4 +97,35 @@ export class CollectionService extends AbstractCachedHttpService<ICollectionRang
...
+          'x-timeout': '180000'
+        }
+      }
+
+    }).pipe(
+      map(({data}) => data),
+    );
+  }
+
+  getCollectionsAggregate(page = 1, collections = []): Observable<any> {
```
vixriihi, line 117: same here

vixriihi (changes requested):
Looks really good :) Few more tweaks and it's perfect

projects/laji/src/app/shared/service/collection.service.ts
MLumme: Typing is done, as is the fix to to getCollectionsAggreagate.

vixriihi (approved):
LGTM

PR #22, opened 2021-01-11
Update to angular 11

vixriihi:


vixriihi:
This needs the be rebased when we decide to include this since migration scripts need to be ran again against the components that have been added after this

olzraiti:
This is a quite massive PR, but if I understood correct it was done with a script provide by Angular? Maybe after the next release we can rerun the script & merge this PR?

vixriihi:
Yes 99.9% were done with the update script :) Only one manual change was needed (return type of slice pipe had changed that was causing an error). Lets see tomorrow if we want to include this too now.

PR #23, opened 2021-01-13
Sidebar in mobile theme page

Hvar83:
Closing the sidebar when any of the nav links are clicked would be desirable on mobile


projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -131,20 +136,24 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
...
-              || (form.options?.useNamedPlaces && (url.match(/\/form\/places/) || url.match(/\/form\/MHL.*\/places/)))
+              || (!this.isDesktopScreen)
             )
           )
         )
       )
-    );
+    ).subscribe(showNav => {
+      this.showNav$ = showNav !== this.openSidebar ? of(showNav) : of(this.openSidebar);
+    });
```
olzraiti, line 156: There seems to be some confusion here about how RXJS works. An observable (`showNav$`) should compose it's output value from piping other observables. Reinitializing the whole observable is going to cause problems (here it is initalized at `ngOnInit` and later on at `checkToggledSidebar`). 

In order to get `showNav$` to output the right values, the proper approach is to turn all the data you need for composing the right value into observables (or if they are observables already like the desktop screen check is, use those) and use those as input observables. So instead of binding the `isDesktopScreen` into a variable, use it as an input observable for `showNav$`, like this:

```
this.showNav$ = combineLatest(this.browserService.lgScreen$, routerEvents$).pipe(
  mergeMap([isDeskopScreen, url]) =>
         form$.pipe(
          map(form =>
            !(
              (!form.options?.useNamedPlaces && url.match(/\/form$/))
              || (form.options?.useNamedPlaces && url.match(/\/places\/MNP\.\d+$/))
              || (url.match(/\/form\/(.*\/)?((JX\.)|(T:))\d+$/))
              || (isDesktopScreen)
            )
          )
        )
  )
)
```

To elaborate why, here are some reasons:
1.You can't in theory be sure that `this.isDesktopScreen` is initialized when you initialize `showNav$`. When using the desktop screen check as an input for the observable, you can be certain that it's initialized.
2. If the screen size changes, updating `this.isDesktopScreen` won't trigger checking new value for `showNav$`.

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -245,4 +255,9 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
   trackByLabel(index, link) {
     return link.label;
   }
+
+  checkToggledSidebar(event) {
+    this.openSidebar = event;
+    this.showNav$ = of(event);
+  }
```
olzraiti, line 262: Is it necessary to know whether the sidebar has been toggled? There could be some reason I'm not seeing. I'm pretty sure this is the reason for the following bug:

1. Go to named places view in desktop mode
2. Close sidebar
3. Click some place
4. Sidebar opens

olzraiti (changes requested):

projects/laji/src/app/+project-form/project-form.component.ts
Hvar83: Ok

projects/laji/src/app/+project-form/project-form.component.ts
Hvar83: That idea to check if sidebar is open/closed was a little confused. But the problem is in sidebar component when it's setting "open" value. As in mobile ShowNav is always false and it never changes, the "toggled" emitter doesn't emit

olzraiti:
This bug is still present:

1. Go to named places view in desktop mode
2. Close sidebar
3. Click some place
4. Sidebar opens (it shouldn't)

This happens because step 3 triggers route event,  and the `subcsriptionShowNav` sets `showNav` to `false`.

In order to fix the bug, the `subscriptionShowNav` needs to know if the sidebar has been closed by the user. I recommend that we make `showNav` an observable `showNav$` like it was before, and make it listen to the `isNavabarToggled` variable. We need to turn `isNavaBarToggled` into an observable, put it to `showNav$`'s input observables and then take it into account when  composing the value of `showNav$`.

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -46,14 +46,19 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
...
+    private router: Router,
+    private browserService: BrowserService
 ) {}
 
   vm$: Observable<ViewModel>;
   formPermissions$: Observable<FormPermission>;
   showNav$: Observable<boolean>;
   isPrintPage$: Observable<boolean>;
   redirectionSubscription: Subscription;
+  isDesktopScreen: boolean;
```
olzraiti, line 58: Remove unused variable

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -46,14 +46,19 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
...
+  isNavabarToggledSubject: Subject<boolean> = new BehaviorSubject<boolean>(false);
+  isNavabarToggled$ = this.isNavabarToggledSubject.asObservable();
+  subscriptionShowNav: Subscription;
```
olzraiti, line 61: 1. Fix typo (Nava -> Nav)
2. Are you sure we can't just use `subscriptionShowNav` directly as an input observable for `showNav$`? if we can't then maybe it'd be cleaner to do the `asObservable()` conversion at directly at the `showNav$` initialization, so we wouldn't have that many instance variables.

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -131,21 +136,21 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
...
+            || (url.match(/\/form\/(.*\/)?((JX\.)|(T:))\d+$/))
+            || (!isDesktopScreen && !isNavbarToggled)
           )
         )
       )
-    );
-
+      )
+    )
+
```
olzraiti, line 153: Looks good! Could you run the indentation formatting tool of your IDE on this block? The indentation at line 141 should be one step deeper.

projects/laji/src/app/+project-form/project-form.component.ts
olzraiti, line 61: Remove unused variable (`subscriptionShowNav`)

olzraiti (changes requested):
Could you run `npm run lint` and fix the errors? This would break the build atm because of the lint errors. There's something funny in your setup maybe since the linting should be run automatically when doing a git commit?

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -136,21 +135,21 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
       startWith(this.router.url)
     );
 
-    this.showNav$ = combineLatest(this.browserService.lgScreen$, routerEvents$, this.isNavabarToggled$).pipe(
-      mergeMap(([isDesktopScreen, url, isNavbarToggled]) => 
+    this.showNav$ = combineLatest(this.browserService.lgScreen$, routerEvents$, this.isNavbarToggledSubject.asObservable()).pipe(
+      mergeMap(([isDesktopScreen, url, isNavbarToggled]) =>
       form$.pipe(
```
Hvar83, line 140: @olzraiti i don't know if like that is good for you

olzraiti: Yes that's good :) One more thing: could you remove the unused `isNavbarToggled$`? Then this would be ready to merge after Mikko accepts the pivotal task. Thanks for the effort!

olzraiti (changes requested):

olzraiti:
Mikko requested on Pivotal this behaviour: https://www.pivotaltracker.com/story/show/176071265/comments/221248189

olzraiti:
I pushed a commit. Could we get a 3rd opinion on this, @vixriihi ?

projects/laji/src/app/+project-form/project-form.component.html
```diff
@@ -1,5 +1,9 @@
 <ng-container *ngIf="vm$ | async as vm; else loader">
-  <lu-sidebar *ngIf="!vm.form.options?.simple && !vm.form.options?.mobile; else simpleContent" [open]="(showNav$ | async)" [noPrint]="isPrintPage$ | async">
+  <lu-sidebar *ngIf="!vm.form.options?.simple && !vm.form.options?.mobile; else simpleContent"
+  [open]="(showNav$ | async)"
+  [noPrint]="isPrintPage$ | async"
```
vixriihi, line 4: hmm can't pinpoint the exact spot but noPrint doesn't seem to [work here](https://176071265.dev.laji.fi/project/MHL.1/form/MHL.1/places/MNP.27299/print) but it's working in [beta](https://beta.laji.fi/project/MHL.1/form/MHL.1/places/MNP.27299/print). (Maybe merge dev branch here to make sure that the error is not caused changes in this branch)

vixriihi (changes requested):
Looks good. Just one thing that should be checked (printing routes seems to be broken here)

projects/laji/src/app/+project-form/project-form.component.html
olzraiti: That doesn't seem to be working in `development` branch. The sidebar and the print button have `no-print` class, and emulating print mode in chrome hides those as it's supposed to. For some reason though, the sidebar & the print button are shown in the generated pdf... I have a vague memory that I've run into this issue before, but can't recall what the issue was.

projects/laji/src/app/+project-form/project-form.component.ts
```diff
@@ -131,15 +133,15 @@ export class ProjectFormComponent implements OnInit, OnDestroy {
...
-    this.showNav$ = routerEvents$.pipe(
-      mergeMap((url: string) =>
+    this.showNav$ = combineLatest(routerEvents$, this.userToggledSidebar$.asObservable()).pipe(
+      mergeMap(([url, userToggledSidebar]) =>
         form$.pipe(
           map(form =>
-            !(
+            userToggledSidebar !== false
+            && !(
               (!form.options?.useNamedPlaces && url.match(/\/form$/))
```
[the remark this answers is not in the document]
olzraiti: Which part of the logic do you mean would be put to route data? The value of `userToggledSidebar` or the whole logical clause for the value of `showNav$`?  Do you mean that the state of the sidebar being open/closed would be a query param?

projects/laji/src/app/+project-form/project-form.component.html
vixriihi: hmm it's working in dev, but this might be something to do with the build process in these task branches.

vixriihi (approved):
LGTM

olzraiti (approved):

PR #29, opened 2021-01-19
Named place linker 175990366

olzraiti:
Allows linking a named place for a document. The button for accessing the linker component is visible on the document viewer and the document form header for documents from forms with named places option when the document doesn't have a named place.

Had to do some refactoring to the named places components, because I used them for the named place linker component (I made the np components prop-controllable, and added a wrapper for the the main named places view that passes the data from the route to the components. I don't know if we should make that into a pattern - reusing components would be much easier if the route data was parsed independently from the component).


projects/laji/src/app/+project-form/form/form.component.ts
```diff
@@ -107,8 +107,12 @@ export class FormComponent implements OnInit {
 
   tryRedirectToSubForm(form: Form.SchemaForm, routeParams: Params): Observable<boolean> {
```
vixriihi, line 108: [just a comment]: This method seems to be called from ngOnInit which is ran rather late in the component initialization that can case small flicker from time to time. I'm wondering if we could move most of this logic to canActivate guard and make the redirect there as needed?

projects/laji/src/app/shared/service/form.service.ts
```diff
@@ -34,14 +35,15 @@ export class FormService {
...
+    private userService: UserService,
+    private translate: TranslateService
   ) {}
 
   static isTmpId(id: string): boolean {
     return !id || (id && id.indexOf(FormService.tmpNs + ':') === 0);
   }
 
-  getForm(formId: string, lang: string): Observable<Form.SchemaForm> {
+  getForm(formId: string, lang = this.currentLang): Observable<Form.SchemaForm> {
```
vixriihi, line 46: better default value could be `this.translate.currentLang` since this.currentLang is set after this

projects/laji/src/app/shared/service/form.service.ts
```diff
@@ -69,7 +71,7 @@ export class FormService {
     return this.jsonFormCache[formId];
   }
 
-  getAllForms(lang: string = this.currentLang): Observable<Form.List[]> {
+  getAllForms(lang = this.currentLang): Observable<Form.List[]> {
```
vixriihi, line 74: `this.translate.currentLang`here also

vixriihi (changes requested):

projects/laji/src/app/+project-form/form/form.component.ts
olzraiti: Good idea - I added a separate task for this https://www.pivotaltracker.com/story/show/176574492

projects/laji/src/app/shared/service/form.service.ts
olzraiti: Good catch, thanks!

vixriihi (approved):

PR #32, opened 2021-01-22
Observation page filters disappear if user switches language [#176553320]

vixriihi:


aorin:
Looks good. There are still some preserveQueryParams left in ts files (like +observation/search-query.service.ts), should they be converted too?

vixriihi:
nice thanks did miss those. They had false value so they would have worked correctly, but it's always good idea to clean up :)

PR #33, opened 2021-01-25
Separate comma vihko forms #168186264

Hvar83:
Split keywords in Vihko document by comma


projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -18,13 +18,15 @@ export enum SpecialTypes {
...
   time = 'time',
+  keywords = 'keywords'
 }
 
 @Injectable()
 export class MappingService {
 
   public static readonly mergeKey = '_merge_';
   public static readonly valueSplitter = ';';
+  public static readonly vihkoFormsSplitters = [';'];
```
vixriihi, line 29: This seems to be no longer used but it could be used in the `mapKeywords`. Maybe rename this to `keywordSplitters = [';', ',']` and use that in the mapKeywords method

projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -18,13 +18,15 @@ export enum SpecialTypes {
   dateTime = 'dateTime',
   date = 'date',
   time = 'time',
+  keywords = 'keywords'
```
vixriihi, line 21: 👍

vixriihi (changes requested):

projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -360,9 +363,16 @@ export class MappingService {
...
+  mapKeywords(value, allowUnMapped) {
+    return value.split(',').map(item => item.trim());
+  }
+
   private _map(value: any, field: IFormField, allowUnMapped = false, convertToArray = true): TUserValueMap|TUserValueMap[]|null {
     if (Array.isArray(value)) {
       value = value.map(val => this._map(val, field, allowUnMapped, false));
+      if (Array.isArray(value[0])) {
+        return value[0];
+      }
```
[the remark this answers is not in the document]
Hvar83: I don't know if i understood well that point...now i'm splitting by [...valueSplitter, ',']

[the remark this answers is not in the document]
vixriihi: If you test this with keyword value `k1;k2,k3`. The current code will will only save `['k1'] ` value to keywords and drops the k2 and k3 values all together.

projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -365,7 +366,7 @@ export class MappingService {
   }
 
   mapKeywords(value) {
-    return value.split(new RegExp([...MappingService.valueSplitter, ','].join('|'), 'g')).map(val => val.trim());
+    return value.split(new RegExp(MappingService.keywordSplitters.join('|'), 'g')).map(val => val.trim());
```
Hvar83, line 369: @vixriihi Related to what you said here "If you test this with keyword value k1;k2,k3. The current code will will only save ['k1'] value to keywords and drops the k2 and k3 values all together." Now i changed and i don't see that mistake

vixriihi (approved):

PR #37, opened 2021-01-29
improve scrollbar accessibility

Blodir:
Completes task #176609954

Scrollbar default width is restored to default. I also did small changes to scrollbar colors to more closely align it with the theming of the site.


vixriihi (approved):
👍

PR #38, opened 2021-01-29
Charts not working individual count #176667352

Hvar83:
Solve problem onlyCount variable. Now calling always false the first time, later taking data from previous result.
Solve problem label ObservationMonthDayChartComponent. It used subscribe to get label chart data, but it had to be observable


projects/laji/src/app/shared-modules/observation-result/observation-month-day-chart/observation-month-day-chart.component.ts
```diff
@@ -365,7 +326,13 @@ export class ObservationMonthDayChartComponent implements OnChanges, OnDestroy,
     this.barChartData = [];
     this.daybarChartData = [];
     this.barChartLabels = [];
-    this.setData(list);
+    this.setData(list).subscribe(
+      () => {
+        this.hasData.emit(this.barChartData.length > 0);
+        this.isNotLoading = false;
```
vixriihi, line 332: double negation here

vixriihi (approved):

PR #41, opened 2021-02-02
Laji UI buttons to overwrite bootstrap buttons

Blodir:
* Laji-ui now implements buttons as mixins (to avoid adding bootstrap dependency to laji-ui)
* Mixins are used to overwrite bootstrap buttons in bs_overwrites.scss (outside of laji-ui, included in laji.scss)
* Loading state is implemented with a css class .lu-loading (remember to also use the html disabled attribute... right now considering a new lu-loading directive that would change both properties, but it has not been implemented yet)

Next up all buttons can be changed to regular html button and anchor elements + bootstrap classes + disabled/loading if needed. After that the original lu-button component can be completely removed.


olzraiti:
Looking good! I hope other people have time to do some testing since this affects the whole site globally.

Some issues I found after some testing:
1. The display and delete buttons are too big on https://176714294.dev.laji.fi/vihko/ownSubmissions page compared to how it used to look  (https://dev.laji.fi/vihko/ownSubmissions)
2. I think buttons with a role should fill the whole button with it's color and use white for content, instead of using white as the filling color and the role color for border & content. E.g. the display & delete buttons on https://176714294.dev.laji.fi/vihko/ownSubmissions and also the document footer buttons (https://176714294.dev.laji.fi/project/JX.519) look off.
3. A disabled button is responsive, it probably shouldn't (at least the "Get place data" button on trip report form responds to mouse hover (visible after adding a gathering and a geometry for the gathering))

This update affects Vihko also. I'd like to hear an opinion at least @mikkohei13 considering Vihko.  I think it's ok if point 2 from above is resolved. Vihko styles look maybe a bit off now (e.g. the boolean switch component isn't consistent with the buttons anymore, date today/yesterday buttons look a bit off)... This could be a good push for making Vihko framework agnostic.

vixriihi:
Code looks ok. This changes the button styling, but I'm guessing it's intentional.

This overriding bs3 styles by adding new styles. In the long run we might want to use the [bs3-sass](https://github.com/twbs/bootstrap-sass) package or update to bs4 so that we'd have a native sass support and then we'd be able to override the original classes/variables?

olzraiti:
I pushed the updated laji-form package to this PR (laji-form buttons aren't primary & not using strong tag by default).

There are some issues with button sizes on for example trip report form (gathering delete button icon isn't centered, unit audio metadata edit button broken). This could be fixed by removing the `laji-form.component.scss` rules and also the bs button custom size in `bs_overwrites.scss`. If the bs buttons used the default bs sizes, they'd be same size as `lu-button` (font size 14). Now the bs buttons actually look bit too large to me (there aren't that many places with bootstrap buttons other than the forms, but for example here they look a bit big https://176714294.dev.laji.fi/vihko/tools/generate).

Blodir:
There are 4 sizes of buttons: btn-lg, btn, btn-sm, btn-xs. I made these correspond to our font-size variables (`$fs-n`). lu-button default is the same as the new default (`.btn`) ie. `$fs-3` which is currently defined as 16px. I also want to note that 16px is a pretty standard reading font size these days (it's even on the small end, bs4 uses 16px, but medium.com uses as large as 21px for leipäteksti!), so that's why I made it the standard button size, as well as home page, about page etc. size.

The reason the buttons look big on pages like `vihko/tools/generate` is that the font-size on that page is 14px.

These are the options we have for adjusting the sizes of buttons:
* reduce font sizes of the `$fs-n` variables so that $fs-3 is 14px
* use a smaller variable for buttons by default (`$fs-2`, which is currently defined as 14px), and use `.btn-lg` for eg. home page buttons
* increase font-size globally to 16px. This is more work, but I think it's the biggest payoff. Some UI elements like observation filters for example could use tuned back to use `$fs-2` (14px) to keep the density of the UI, but by default text content would be `$fs-3` (16px) which is much more readable.
* switch to using `.btn-sm` on any button that looks too big (it's already using 14px text!)

olzraiti:
In this PR, `lu-button` size is 14px by default and the bootstrap buttons are 16px. In `dev` branch, bootstrap buttons are also 14px. I think we should keep them at 14px, because we should make the minimum changes to avoid scope creeping. Changing the font sizes to be larger in general is a separate issue and out of this PR's scope.

Bootstrap `xs` and `sm` buttons use the same font size, the `xs` button just has smaller padding. So how about these font sizes for bootstrap:

* `btn-xs`: `fs-1`
* `btn-sm`: `fs-1`
* `btn`: `fs-2`
* `btn-lg`: `fs-3`

This way the `lu-button` and the bootstrap button (without any size modifier -classes) would be the same size, and we wouldn't need to change any font sizes on the page globally.

Blodir:
I think that's fair, at least so we can get this PR merged. Also:
* `btn-xs` has been completely removed in BS4, so we could actually get rid of this completely in preparation for possible upgrade anyway
* `btn-lg` is too large to be useful (like it is in default BS3 as well), so it makes sense to reduce it to `fs-3`

After testing it a bit I also noticed that a 14px button doesn't look too bad at all against 16px bodycopy in general (and it obviously works with current 14px too).

olzraiti:
Aight, I think we can wrap this up now, thanks :)

Getting ready for the bs4 upgrade sounds good. `btn-lg` isn't used anywhere, so that won't be a problem. `btn-xs` on the other hand is used quite a lot:

```
$ grep btn-xs -r projects | wc -l
31
```

Judging from the amount of how much we currently rely on the `xs` sized button, we might have a justified reason for using an extra small button. Or not - but removing them will require some redesigning for quite many components. Redesigning our UI to work without extra small buttons might be beneficial in any case, but for the bs4 migration we could copy the bs3 `xs` button styles to our code base just to make the transition technically smoother. We can discuss this further when we decide to do the migration.

olzraiti:
It seems that I can't approve this PR - probably because I committed some changes myself. Could you @vixriihi approve since you already had a look on this?

Blodir:
I think `btn-sm` is probably small enough to replace btn-xs completely or `btn-sm` can be made the same size as old `btn-xs`. Atm there's only a couple of pixels difference so they are pretty much interchangeable. 
image

package-lock.json
```diff
@@ -16508,11 +16507,6 @@
         }
       }
     },
-    "react-bootstrap-switch": {
```
vixriihi, line 16511: Seems like removing this breaks the switch [here](https://176714294.dev.laji.fi/taxon/list?onlyFinnish=true)

vixriihi (changes requested):
Looks good. One small thing needs tweaking and then this would be ready

package-lock.json
olzraiti: Good catch, fixed. It's a bit worrisome that changing laji-form styles breaks laji.fi styles... (I removed the bootstrap-switch from laji-form in this PR, and apparently laji.fi wasn't taking care of loading the styles itself. Which makes sense though, because we don't want to include the styles twice)

vixriihi: completely agree. Style dependency should have been added to laji.fi side when the switch was added

vixriihi (approved):

PR #43, opened 2021-02-02
Custom dialogs #176741042

olzraiti:
Implements `DialogService` `.confirm()` and `.prompt()` with bootstrap modals instead of browser dialogs.

[Demo](https://176741042.dev.laji.fi/)


projects/laji/src/app/shared/service/dialog.service.ts
```diff
@@ -1,29 +1,53 @@
...
+      subject.next(prompt ? null : false);
+      subject.complete();
+    };
+    modalRef.content.cancel.subscribe(cancel);
+    modalRef.onHide.subscribe(cancel);
+    modalRef.content.cdr.markForCheck();
+    return prompt
+      ? subject as Subject<string | null>
+      : subject as Subject<boolean>;
+  }
```
vixriihi, line 52: You could leverage initialState option in modalService and instead of returning subject return an observable that has the desired value. So that this would become Something like this:
```
    const modalRef = this.modalService.show(ConfirmComponent, {
      backdrop: 'static',
      initialState: {message, prompt, promptValue: promptDefault ?? ''}
    });

    return modalRef.onHidden.pipe(
      map(() => modalRef.content.value)
    ); 
```
Then add the `value` property to confirmComponent and use `constructor(public bsModalRef: BsModalRef) {}` in there to close the modal when needed and set the value to what is should return here

vixriihi (changes requested):

PR #46, opened 2021-02-04
convert multiple area codes - [176759772]

Hvar83:
When multiple municipalities, areas were not converted to name

### DIFF
--- a/projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
+++ b/projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
@@ -34,9 +34,9 @@
 <alert *ngIf="editingOldWarning" type="warning">{{ 'haseka.editingOldDocumentWarning' | translate: {date: formatDate(formData.dateCreated)} }}</alert>
 <div class="row" *ngIf="namedPlace" style="margin-top: 10px">
   <div class="col-sm-12">
-    <h3><span *ngFor="let field of namedPlaceHeader; let last = last">
+    <h3><span *ngFor="let field of namedPlaceHeader; let i = index">
       <ng-container *ngIf="namedPlace[field] | area | values as val">
-        {{ val }}{{ !last && val ? ', ' : '' }}
+       {{ val }}{{ namedPlace[namedPlaceHeader[i + 1]] ? ', ' : '' }}
       </ng-container>
     </span></h3>
   </div>
--- a/projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
+++ b/projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
@@ -34,7 +34,6 @@
 <alert *ngIf="editingOldWarning" type="warning">{{ 'haseka.editingOldDocumentWarning' | translate: {date: formatDate(formData.dateCreated)} }}</alert>
 <div class="row" *ngIf="namedPlace" style="margin-top: 10px">
   <div class="col-sm-12">
-    <!--<h3><span *ngFor="let field of namedPlaceHeader; let last = last">{{ field | async | area }}{{ !last ? ', ' : '' }}</span></h3>-->
     <h3><span *ngFor="let field of namedPlaceHeader; let last = last">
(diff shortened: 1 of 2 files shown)


olzraiti:
Thanks for tackling this! Seems to work otherwise, but for the municipalities don't have a space between them (e.g. on https://176759772.dev.laji.fi/project/MHL.45/form/MHL.45/places/MNP.38120 it reads "test, Eckerö,Helsinki", when it should read "test, Eckerö, Helsinki").

To fix that, add `values` pipe after the `area` pipe.

However, this implementation is contradicting with the `getNamedPlaceHeader()` logic. The `getNamedPlaceHeader()` is returning observables of converted strings, the idea being that it's extendable if other async labels than municipalities would be needed in the future. Now the change in the template you made assumes that the only type of values needing label conversion is municipalities. The original `getNamedPlaceHeader()` might have been over engineered, so IMO we can go your way, but could you move the logic entirely to the template:

In the template:

```
    <h3><span *ngFor="let field of namedPlaceHeader; let last = last">
      <ng-container *ngIf="namedPlace[field] | area | values as val">
        {{ val }}{{ !last ? ', ' : '' }}
      </ng-container>
    </span></h3>
```

Update `getNamedPlaceHeader()` (`document-form-header.component.ts`):
```
  getNamedPlaceHeader(): string[] {
    if (!this.form || !this._namedPlace) {
      return [];
    }
    return this.form.options?.namedPlaceOptions?.headerFields || ['alternativeIDs', 'name', 'municipality'];
  }
```

projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.html
```diff
@@ -34,7 +34,12 @@ <h3 *ngIf="displayTitle">{{ form.title }}</h3>
 <alert *ngIf="editingOldWarning" type="warning">{{ 'haseka.editingOldDocumentWarning' | translate: {date: formatDate(formData.dateCreated)} }}</alert>
 <div class="row" *ngIf="namedPlace" style="margin-top: 10px">
   <div class="col-sm-12">
-    <h3><span *ngFor="let field of namedPlaceHeader; let last = last">{{ field | async | area }}{{ !last ? ', ' : '' }}</span></h3>
+    <!--<h3><span *ngFor="let field of namedPlaceHeader; let last = last">{{ field | async | area }}{{ !last ? ', ' : '' }}</span></h3>-->
```
olzraiti: rm commented line

olzraiti (changes requested):

projects/laji/src/app/shared-modules/laji-form/document-form-header/document-form-header.component.ts
```diff
@@ -101,30 +101,10 @@ export class DocumentFormHeaderComponent implements OnInit, OnChanges, OnDestroy
     return moment(date).format('DD.MM.YYYY');
   }
 
-  getNamedPlaceHeader(): Observable<string>[] {
+  getNamedPlaceHeader(): string[] {
     if (!this.form || !this._namedPlace) {
       return [];
     }
-    const headerFields = this.form.options?.namedPlaceOptions?.headerFields || ['alternativeIDs', 'name', 'municipality'];
```
Hvar83, line 108: @olzraiti Changes you asked are here. I'm leaving now, but if there is some problem, i will fix tomorrow morning

olzraiti: Looks good, thanks! I'll wait for the feature demo to build, test and merge

olzraiti:
If the last field is empty, an excess comma separator is outputted. To fix that use:

```
{{ val }}{{ !last && val ? ', ' : '' }}
```

olzraiti:
Sorry that was wrong, I didn't check that properly. The bug still exists... checking `val` doesn't help, because the problem is that we don't know if the next field has a value when deciding whether to render a comma. You can check the bug by reserving some line transect place and reporting a document for it.

I can think of two options (just suggestions, implement as you wish :)

1. Create a filter pipe so that you could filter the fields that don't have value
2. Put the logic of getting the value array in the component code after all, so you can filter the values there.

PR #48, opened 2021-02-05
Let user move between previous or next observation(image) and user other shortcut in identify page

Hvar83:
It's working, and i modified many components to make it possible.
i guess it's not high priority, we can comment after other tasks are published. but i think it needs a little check more to make the component <document-annotation> usable in all components.


projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -333,9 +335,20 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
...
+                (element.media || []).forEach(image => {
+                  const i = image.fullURL === this.highlight;
+                  if (i) {
+                    return this.imgExist = true;
+                  }
+                });
+              });
             }
           }
         });
```
olzraiti, line 354: `this.unitExists` and `this.imgExists` could be combined into a single variable. How about `unitOrImgExists`?  Also, I think using `some` would be more understandable here instead of a `forEach` since you are basically searching for a single occurrence of a unit or an img. It'd also be more performant, since you can stop the search right when you find a unit or an img (returning `true` doesn't stop the `forEach` in plain JS, though in jQuery it does :) Could also have less indentations.

Proposal (replaces lines 335-355):

```
      this.unitOrImgExists = this.document.gatherings?.some(({units = []}) =>
          this.identifying
            ? units.some(element =>
              (element.media || []).some(image => image.fullURL === this.highlight)
            )
            : units.some(unit => unit.unitId === this.highlight)
      );
```

projects/laji/src/app/shared-modules/document-viewer/images/images.component.ts
```diff
@@ -50,13 +51,31 @@ export class ImagesComponent implements OnChanges {
...
+            );
+          } else {
+            let imgUnitId;
+            gathering.units.forEach(element => {
+              (element.media || []).forEach(image => {
+                if (image.fullURL === this.highlight) {
+                  imgUnitId = element.unitId;
+                }
+              });
+            });
```
olzraiti, line 70: Unnecessary `forEach` iteration, would be more performant with `find` (something like `const imgUnitId = gatherings.units.find(...`). The same goes for the inner `forEach`

olzraiti (changes requested):

projects/laji/src/app/+theme/identify/identify.component.html
```diff
@@ -37,7 +37,18 @@ <h1>{{ 'navigation.identify' | translate }}</h1>
   <div class="row">
     <div class="col-sm-12 image-container">
       <ng-container *ngIf="query">
-        <laji-gallery [showPopover]="true" [shortcut]="false" [view]="'full'" [sort]="['document.firstLoadDate DESC']" [showViewSwitch]="true" [query]="query" [eventOnImageClick]="true" (selected)="onImageSelect($event)"></laji-gallery>
+        <laji-gallery
```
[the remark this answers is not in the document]
olzraiti: Sorry, there are no empty lines :) I understood the github diff wrong.

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
olzraiti: Actually even shorter:

```
this.unitOrImgExists = this.document.gatherings?.some(({units}) =>
  (units || []).some(unit =>
    this.identifying
      ? (unit.media || []).some(image => image.fullURL === this.highlight)
      : unit.unitId === this.highlight
  )
);
```

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -283,8 +282,7 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
     this.cd.detectChanges();
     this.hasDoc = found;
     this.hasEditors = false;
-    this.unitExist = false;
-    this.imgExist = false;
+    this.unitOrImgExists = false;
```
olzraiti, line 285: Unnecessary initialization, the value is computed later.

projects/laji/src/app/shared-modules/document-viewer/images/images.component.ts
```diff
@@ -50,13 +51,31 @@ export class ImagesComponent implements OnChanges {
...
+            gathering.units.filter(unit => unit.unitId === imgUnitId).map(
```
vixriihi, line 71: This repeats almost the same logic twice. Instead you could do this kind of logic:
(highlight can now be unit id or image fullURL)
find the unit that has the highlight id or image that has the highlighted full url
and then add that units media to this.unitImages list
something like this:
``` 
if (!gathering.units) {
  return;
}
const unit = gathering.units.find(u => u.id === this.highlight || (u.media || []).some(media => media.fullURL === this.highlight));
if (unit?.media) {
  this.unitImages.concat(unit.media);
}

**Review by vixriihi — CHANGES_REQUESTED** [review]

File: projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts [hunk 6]
```diff
@@ -373,7 +369,11 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
   }
 
   setIndexPagination() {
-    return this.result.findIndex(i => i.unit.unitId === this.highlight);
+    if (!this.identifying) {
+      return this.result.findIndex(i => (i.unit.unitId === this.highlight));
+    } else {
+      return this.result.findIndex(i => (i.fullURL === this.highlight));
```
vixriihi, line 375: this part doesn't need to use `this.identifying` since this could be written as:
```
return this.result.findIndex(i => (i.fullURL === this.highlight || i.unit?.unitId === this.highlight))
```
ps the name of this method is also little misleading sine it's not setting anything but returning the index

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -435,8 +435,8 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
 
   private move() {
     this.document = this.result[this.indexPagination].document;
-    this.uri = this.result[this.indexPagination].document.documentId;
-    this.highlight = this.result[this.indexPagination].unit.unitId;
+    this.uri = this.identifying ? this.result[this.indexPagination].documentId : this.result[this.indexPagination].document?.documentId;
+    this.highlight = this.identifying ? this.result[this.indexPagination].fullURL : this.result[this.indexPagination].unit?.unitId;
```
vixriihi, line 439: this.identifying is not needed in this method either

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -330,16 +329,13 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
...
-            const i = gathering.units.find(unit => unit.unitId === this.highlight);
-            if (i) {
-              return this.unitExist = true;
-            }
-          }
-        });
-      }
+      this.unitOrImgExists = this.document.gatherings?.some(({units}) =>
+        (units || []).some(unit =>
+          this.identifying
```
vixriihi, line 334: `this.identifying` is not needed here either. Same kind of solution applies to this that what was already implemented to ImagesComponent. (I'm not sure how the identifying input should be thought in the context of annotation viewer and that's why I'd try to remove that (or possibly rename it to something more descriptive in this context))

vixriihi (changes requested):

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -435,8 +430,9 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
 
   private move() {
     this.document = this.result[this.indexPagination].document;
-    this.uri = this.identifying ? this.result[this.indexPagination].documentId : this.result[this.indexPagination].document?.documentId;
-    this.highlight = this.identifying ? this.result[this.indexPagination].fullURL : this.result[this.indexPagination].unit?.unitId;
+    this.result.findIndex(i => (i.fullURL === this.highlight || i.unit?.unitId === this.highlight));
```
vixriihi, line 433: This might be just accidental pasting?

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -435,8 +430,9 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
 
   private move() {
     this.document = this.result[this.indexPagination].document;
```
vixriihi, line 432: This line is little strange since I guess that `this.result` can be a list of documents or images now?

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -435,8 +430,9 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
...
-    this.uri = this.identifying ? this.result[this.indexPagination].documentId : this.result[this.indexPagination].document?.documentId;
-    this.highlight = this.identifying ? this.result[this.indexPagination].fullURL : this.result[this.indexPagination].unit?.unitId;
+    this.result.findIndex(i => (i.fullURL === this.highlight || i.unit?.unitId === this.highlight));
+    this.uri = (this.result[this.indexPagination].documentId || this.result[this.indexPagination].document?.documentId);
+    this.highlight = (this.result[this.indexPagination].fullURL || this.result[this.indexPagination].unit?.unitId);
```
vixriihi, line 435: How about:
```
const documentOrImage = this.result[this.indexPagination].document || this.result[this.indexPagination]
```
then bellow use something like this:
```
this.uri = documentOrImage.documentId
this.highlight = documentOrImage.fullURL || documentOrImage.unit?.unitId

**Review by vixriihi — CHANGES_REQUESTED** [review]

File: projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts (hunk 11 above)
Comment (Hvar83) [reply projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts]: @vixriihi i made a solution a little bit different than yours. But when you can check the code we can discuss if it's not good

File: projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts [hunk 12]
```diff
@@ -429,10 +429,9 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
   }
 
   private move() {
-    this.document = this.result[this.indexPagination].document;
-    this.result.findIndex(i => (i.fullURL === this.highlight || i.unit?.unitId === this.highlight));
-    this.uri = (this.result[this.indexPagination].documentId || this.result[this.indexPagination].document?.documentId);
-    this.highlight = (this.result[this.indexPagination].fullURL || this.result[this.indexPagination].unit?.unitId);
+    this.documentOrImage = this.result[this.indexPagination];
```
vixriihi, line 432: good if this can be done :)

Now though the documentOrImage would hold when result is a list of documents a value that would be something like this `{document: {documentId: 'JA.1', ...}}` Previously this picked `.document` from the results with the wanted index so that it would have hold something like this: `{documentId: 'JA.1', ...}` .  Can you confirm that this is correct and I'll test this more thoroughly

vixriihi (changes requested):

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
Hvar83: @vixriihi i can confirm that it's working at least in Observation/annotations and identify pages. Result in observation/annotations is an object that contains other object "document" and in indentify is a single object which contains properties "documentId" and "FullURL"

vixriihi: `this.document` based on the view should only hold document information I added a local variable to this method so that it would not reuse the `this.document` to something that it was not indented. Reason why it works is that right after in this method it calls `updateDocument()` which in turn updates that variable back to document

vixriihi (approved):
LGTM

PR #54, opened 2021-02-12
fix IUCN pages halting [#176918563]

vixriihi:


projects/laji/src/app/shared/pipe/label.pipe.ts
```diff
@@ -18,100 +18,70 @@ type labelType = 'qname'|'fullUri'|'warehouse'|'withKey'|'emptyWhenMissing';
...
-    if (Array.isArray(value)) {
-      return value.map(v => this.transform(v, type, key));
-    }
-    if (value && key) {
-      value = value[key];
-    }
-    if (!value || typeof value !== 'string' || value.length === 0 ||
-       (type === 'fullUri' && value.indexOf('http') !== 0)) {
+  transform<T extends string|string[]>(value: T, type?: labelType): T {
+    if (!value || (typeof value !== 'string' && !Array.isArray(value)) || value.length === 0) {
```
olzraiti, line 32: Is this check necessary when `T` is already typed to `string|string[]`? The check could be simply:

```if (!value.length) {```

File: projects/laji/src/app/shared/pipe/pluck.pipe.ts [hunk 2]
```diff
@@ -0,0 +1,15 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  name: 'pluck'
+})
+export class PluckPipe implements PipeTransform {
+
+  transform<T extends any, K extends keyof T>(value: T|T[], key: K): T[K] | Array<T[K]> {
+    if (Array.isArray(value)) {
+      return value.map(v => v?.[key]);
```
olzraiti, line 10: Unnecessary `?.` operator (sorry for nagging, but it's good practice to try to keep compiled code short :)

olzraiti (changes requested):
Looking good! Just a few very minor naggings

projects/laji/src/app/shared/pipe/label.pipe.ts
```diff
@@ -18,100 +18,70 @@ type labelType = 'qname'|'fullUri'|'warehouse'|'withKey'|'emptyWhenMissing';
...
+          of(key);
         break;
       case 'withKey':
         obs = this.triplestoreLabelService.get(key, this.translate.currentLang).pipe(
           map(value => value !== key ? (value || '') + ' (' + key + ')' : value));
         break;
       default:
-        obs = this.triplestoreLabelService.get(key, this.translate.currentLang);
+        obs = this.triplestoreLabelService.get(key, this.translate.currentLang).pipe(
+          map((res: string) => res && res !== key ? res : (type === 'emptyWhenMissing' ? '' : key))
```
olzraiti, line 81: Could we split this to multiple lines? Multiple ternary operators in a single line is a bit hard to read.
E.g. 

```
res && res !== key
  ? res
  : type === 'emptyWhenMissing'
    ? ''
    : key
```

projects/laji/src/app/shared/pipe/pluck.pipe.ts
vixriihi: No if the value is `['foo', null]` with out ?. that would cause an error

projects/laji/src/app/shared/pipe/label.pipe.ts
vixriihi: That allows typing to be so that if value = string the output will be type string, and if value is string[] then the output would be value string[] so this will prevent the case where string input and output would be string[], but method overloading might make this little pit more readable :)

projects/laji/src/app/shared/pipe/label.pipe.ts
vixriihi: sounds good  👍

projects/laji/src/app/shared/pipe/pluck.pipe.ts
olzraiti: Indeed... Sorry, I misread here that `v` would be the array :)

projects/laji/src/app/shared/pipe/label.pipe.ts
olzraiti: Hmm I'm not following... Can you give me an example for a `value` (that is of type `string|string[]`) that behaves differently for these checks:

```
if (!value.length) {
```

```
if (!value || (typeof value !== 'string' && !Array.isArray(value)) || value.length === 0) {
```

vixriihi: Within the method there is nothing that would behave differently. Difference is that the method is returning type T so if you'd use it on a value that is of type string it would keep the return of this function as a string and not return type string|string[] for the code that's using it :)

More about those can be found here https://www.typescriptlang.org/docs/handbook/generics.html

vixriihi: ps. yesterday just noticed that there is a handy strictTemplates option in tsconfig that I'm refactoring our code to use 😄

olzraiti: Ok, thanks for the explanation! :)

PR #56, opened 2021-02-12
Protax improvements 176179408

aorin:
Improvements to protax page. Testable [here](https://176179408.dev.laji.fi/theme/protax).


projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
```diff
@@ -35,6 +38,11 @@ export class ProtaxFormComponent {
   }
 
   submitForm() {
+    if (this.probabilityThreshold == null || this.probabilityThreshold < 0 || this.probabilityThreshold > 1) {
```
vixriihi, line 41: `this.probabilityThreshold == null` could be replaced with just `!this.probabilityThreshold` or if you need to check the null case specifically you need to use `this.probabilityThreshold === null`

vixriihi (changes requested):
small tweak suggestions 👍

projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
aorin: `!this.probabilityThreshold` would not work with 0 so I used `this.probabilityThreshold == null` to check that it isn't undefined or null

aorin:
@vixriihi now the alerts use dialog service and ngif repetition is fixed

projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
vixriihi: true 👍

vixriihi (approved):
LGTM. Should this be included with the next release or is it enough to go in the one after?

PR #59, opened 2021-02-18
Kerttu audio viewer improvements

aorin:
Improvements to Kerttu audio viewer (testable [here](https://176883476.dev.laji.fi/theme/kerttu/letters).

New features:
- recording can be played continuously with a loop button
- audio is normalized so that volume wouldn't change so much between recordings
- warning that the page doesn't work if mute switch is turned on is shown with ios devices
- audio viewer shouldn't get stuck with ios devices anymore

I also did some refactoring (especially to audio player service) so this became quite a big task.


projects/laji/src/app/+theme/kerttu/directive/requires-audio-support.directive.ts
```diff
@@ -18,11 +23,21 @@ export class RequiresAudioSupportDirective implements OnInit {
...
+      }
       this.viewContainer.createEmbeddedView(this.templateRef);
     }
   }
 
   private webAudioApiIsSupported() {
     return !!this.audioService.audioContext;
   }
+
+  private isIOS() {
```
vixriihi, line 38: This could be moved to PlatformService where we already have few other checks like this awailable (same for the webAudioApiIsSupported method)

vixriihi (approved):
looks good, small suggestion.

PR #65, opened 2021-03-02
fix shortcut arrow keys in Annotation and Images modal

Hvar83:
shortcut didn't work well in annotation modal after open image modal. Same happens when user open images in taxon images tab.
Also component image-modal-overlay still have previous array of images when user move to annotation, so that generated some problem to shortcuts too


olzraiti (approved):

vixriihi:
this seems to be a bug fix so merging this 👍

PR #70, opened 2021-03-05
restyle pill-list

Blodir:
Before:
image

After:
image


projects/laji/src/app/shared-modules/pill-list/pill-list.component.scss
```diff
@@ -0,0 +1,26 @@
...
+
+  box-shadow: 0 1px 2px 0 rgba(#000, .2);
+
+  .glyphicon-remove {
+    margin-left: $sp-2;
+    font-size: 10px;
+    padding: 3px;
+    color: $success-4;
+    background-color: $success-6;
+    border-radius: 1000px;
```
vixriihi, line 20: `border-radius: 100%` would little nicer

projects/laji/src/app/shared-modules/pill-list/pill-list.component.scss
```diff
@@ -0,0 +1,26 @@
...
+.label {
+  display: inline-block;
+  margin-top: $sp-1;
+  margin-right: $sp-1;
+
+  font-size: $fs-1;
+  color: $success-1;
+  background-color: $success-5;
+
+  box-shadow: 0 1px 2px 0 rgba(#000, .2);
```
vixriihi, line 12: I'm wondering if we should also add a global class like `elevation-1`, `elevation-2` etc so that we wouldn't end up using different styles in different places

vixriihi (approved):
looks good 👍 Few comments for consideration in the style

projects/laji/src/app/shared-modules/pill-list/pill-list.component.scss
Blodir: Hmm in this case either px or % units work just fine, because the height and width are equal. `border-radius` actually works quite differently with px and % units and `border-radius: 1000px;` is a standard instruction for styling rectangles with fully rounded corners (% units don't work for this). In this case maybe % units would be more intuitive 🤔

projects/laji/src/app/shared-modules/pill-list/pill-list.component.scss
Blodir: I've also thought about this, but there's a problem with it: Box-shadow can't be broken down into its component properties. This would be required, because elements with a strong background-color should have colored box-shadows (ironically I forgot to do it here, fixing it).

vixriihi: ok yes first comment was based on the black border color and if the color needs to be alternated then having global class would become a problematic :) Mixing maybe... end goal would be to find some good way to keep the elevation effects uniform.

PR #73, opened 2021-03-08
Obs search open filter fully something selected #177183600

Hvar83:
If a filter is checked, and user refresh page or open a the page by paste the link, show also all other options when filter is closed


vixriihi (changes requested):
Task talks about having the global select [functionality changed](https://www.pivotaltracker.com/story/show/177183600).  But at least in the country select is closed https://177183600.dev.laji.fi/observation/list?countryId=ML.206

projects/laji/src/app/shared-modules/search-filters/three-state-multi-switch/three-state-multi-switch.component.html
```diff
@@ -1,9 +1,9 @@
 <div class="lj-select" [ngClass]="{'lj-close': open, 'lj-open': !open}">
   <strong *ngIf="title !== undefined && title !== ''" class="lj-title link" (click)="toggle($event)">{{ title }}<laji-info *ngIf="info" [html]="info"></laji-info></strong>
   <ng-container *ngIf="options$ | async; let options">
-    <div [hidden]="open">
+    <div [hidden]="open || (options | filterOptions:'id':falseValue:trueValue).length === 0">
```
vixriihi, line 4: instead of adding this logic to view you could add it in the controller so that open property would reflect the desired value

projects/laji/src/app/shared-modules/select/select/select.component.html
```diff
@@ -43,13 +45,13 @@
 </div>
 
 <ng-template #select>
-  <ng-container *ngIf="!open">
-    <span class="lj-container" *ngFor="let option of selectedOptions; trackBy: track">
-      <label class="lj-item selected" (click)="labelClick($event)">
-        <laji-checkbox [value]="true" (valueChange)="remove(option.id, $event)"></laji-checkbox> {{ option.value }}
+  <ng-container *ngIf="!open && selectedOptions.length > 0">
```
vixriihi, line 48: This logic would also be better within the controller

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -103,7 +102,7 @@ export class SelectComponent<T extends idType|SelectOption = string> implements
         this.selected = [...this.selected, id] as T[];
       }
     } else {
-      this.selected = isBasic ? [option] as T[] : [id] as T[];
+      this.selected = isBasic ? [option.id] as T[] : [id] as T[];
```
vixriihi, line 105: It seems like there has been a bug in the original row, but now it would return same kind of result on both cases. Most likely the correct form should be `this.selected = isBasic ? [id] as T[] : [option] as T[];` (Idea is to make this.selected to match the type that was given in the @Input)

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/select/select.component.ts
Hvar83: This wasn't update with dev version, so this issue still was there

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -301,6 +302,8 @@ export class SelectSubcategoriesComponent implements OnChanges {
       });
     }
 
+    this.open = (!this.open && Object.keys(this.selectedOptions).length > 0) ? true : this.open;
```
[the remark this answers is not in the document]
Hvar83: I changed the code in the commit 5c33ca9. If i go inside the method "loopInsideOptions", it means that there is some option (global or subCategory) checked. So no need to check if Object.keys(this.selectedOptions).length > 0 because it will be anyway always true. So the variable open starts in false, but if i call "loopInsideOptions" i will turn it to true.

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -302,7 +302,7 @@ export class SelectSubcategoriesComponent implements OnChanges {
       });
     }
 
-    this.open = (!this.open && Object.keys(this.selectedOptions).length > 0) ? true : this.open;
+    this.open = true;
```
vixriihi, line 305: if this need to be always true. I would remove this line all together and change the line `
  [open]="open"` to `
  [open]="true"` in `metadata-select-with-subcategories.component.html`

vixriihi (changes requested):

vixriihi:
small tweak 👍

vixriihi (approved):

PR #76, opened 2021-03-15
Fixes rows disappear from namedplace select table [#177301350]

vixriihi:
Testable at https://177301350.dev.laji.fi/


olzraiti (approved):
:+1:

PR #77, opened 2021-03-15
Disabled form [#177161689]

olzraiti:
If a form has `MHL.disabled` in the form options, prevent users except ICT admins from accessing the form. This is just the frontend solution, the backend solution is implemented in LajiAPI.

Testable at https://177161689.dev.laji.fi/save-observations

Added a disabled form MHL.90 for testing, which is visible in the survey category.


projects/laji/e2e/src/+project-form/project-form.e2e-spec.ts
```diff
@@ -108,6 +109,13 @@ describe('Project form', () =>  {
         done();
       });
     });
+
+    it('disabled form is displayed as disabled', async (done) => {
+      await projectFormPage.navigateTo(FORM_DISABLED);
+      expect(await projectFormPage.$disabled.isDisplayed()).toBe(true);
```
vixriihi, line 115: hmm this seems to be navigating to address [/project/MHL.90](https://177161689.dev.laji.fi/project/MHL.90). When I go there I do not see the dialog. Should I?

vixriihi (changes requested):
Small tweak 👍 and one question :)

projects/laji/e2e/src/+project-form/project-form.e2e-spec.ts
olzraiti: Did you user your ICT admin account? ICT admins can access disabled forms, regular users/form admins cannot.

vixriihi: That would have been my second guess :) Yes was using ict admin now that tested again with another count dialog is clearly visible to login user and guests 👍

vixriihi (approved):
LGTM 👍

PR #78, opened 2021-03-19
Laji-form hacks removed that have fixed some issues with prod build

olzraiti:
This can be tested only in beta env, so I'll merge this right away, test it and revert the changes if they don't work.


olzraiti:
Everything seems to work, no need to revert.

PR #79, opened 2021-03-23
Mac excel import adds time when importing a field that's date formatted date 177379077

vixriihi:
Fix issue where on Mac cell with only date would also have time in it and fix a issue that would replace value TEST in dateOptionalTime field with value ES


projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -327,17 +327,13 @@ export class MappingService {
...
     } else if (value instanceof Date) {
-      if (value.getHours() === 23 && value.getMinutes() === 59 && value.getSeconds() === 11) {
-        const tmpDate = new Date(value);
-        tmpDate.setMinutes(value.getMinutes() - (value.getTimezoneOffset() - 1));
-        return tmpDate.toISOString().substr(0, 10);
+      if (
+        this.matchTime(value, 0, 0, 0) || // Excel from Mac
+        this.matchTime(value, 23, 59, 11) // Linux & Windows
+      ) {
+        return this.getDate(value);
```
olzraiti, line 336: What's special about 11 seconds past 23:59? Why not always parse just the date from a value of type `Date`?

vixriihi: you have to ask that from the excel 😂 But that's the time that exceljs gives when only date is inputted to the field. If the field has date type then we can use the `YYYY-MM-DD` format (and we are), but this field here has a type of dateOptionalTime meaning that need to somehow differentiate the input that has date from dateTime. I know that this is not optimal and I'm open for other solutions 😃

olzraiti: Right... :D

olzraiti (approved):

PR #80, opened 2021-03-24
Show location on kerttu audio viewer 177222224

aorin:
Testable [here](https://177222224.dev.laji.fi/theme/kerttu/letters). Bottom right corner of the audio viewer shows time and location information of the recording (if available).


projects/laji/src/app/+theme/kerttu/audio-viewer/audio-info/audio-info.component.ts
```diff
@@ -0,0 +1,31 @@
+import {Component, OnInit, ChangeDetectionStrategy, ViewChild, TemplateRef, Input} from '@angular/core';
+import {IAudio} from '../../models';
+import {BsModalRef, BsModalService} from 'ngx-bootstrap/modal';
```
olzraiti, line 3: Wrap imports with spaces for consistency

projects/laji/src/app/+theme/kerttu/audio-viewer/audio-info/audio-info-map/audio-info-map.component.ts
```diff
@@ -0,0 +1,51 @@
...
+    zoomToData: { maxZoom: 13 }
+  };
+
+  data: any;
+
+  @Input() set geometry(geometry: GeoJSON.Geometry) {
+    this.data = this.getData(geometry);
+  }
+
+  constructor() { }
```
olzraiti, line 29: Remove unused constructor

projects/laji/src/app/+theme/kerttu/audio-viewer/audio-info/audio-info.component.ts
```diff
@@ -0,0 +1,31 @@
...
+  ) { }
+
+  openModal() {
+    this.modal = this.modalService.show(this.mapModal, {class: 'modal-md'});
+  }
+
+  closeModal() {
+    if (this.modal) {
+      this.modal.hide();
+    }
```
olzraiti, line 29: `this.modal?.hide()` would be shorter

projects/laji/src/app/+theme/kerttu/kerttu-letter-annotation/kerttu-letter-annotation.component.ts
```diff
@@ -165,7 +165,7 @@ export class KerttuLetterAnnotationComponent implements OnInit, OnDestroy {
       .pipe(
         switchMap((result) => {
           const candidate = result.candidate;
-          return this.audioService.getAudioBuffer(candidate.recording).pipe(map(() => candidate));
+          return this.audioService.getAudioBuffer(candidate.audio.url).pipe(map(() => candidate));
```
olzraiti, line 168: Seems like the result from the `this.audioService.getAudioBuffer` is never used, as the `map()` just returns the `candidate`?

projects/laji/src/app/+theme/kerttu/models.ts
```diff
@@ -1,8 +1,16 @@
 import {Taxonomy} from '../../shared/model/Taxonomy';
+import {GeoJSON} from 'geojson';
```
olzraiti, line 2: Wrap imports with spaces for consistency

olzraiti (changes requested):

projects/laji/src/app/+theme/kerttu/kerttu-letter-annotation/kerttu-letter-annotation.component.ts
aorin: It isn't used at that point but it loads the buffer to audio service cache. It makes changing of next candidate faster when the audio is already loaded.

olzraiti: Hmm... could you just do the same with tap to make it more readable:

```
this.nextLetterCandidate$ = this.kerttuApi.getNextLetterCandidate(this.userService.getToken(), templateId, candidateId)
  .pipe(
    map((result) => result.candidate),
    tap(candidate => this.audioService.getAudioBuffer(candidate.audio.url).pipe(take(1)).subscribe()),
    share()
);
```

The reasoning being, that loading the audio buffer to the cache is a side effect, and the `nextLetterCandidate$` value comes from the `map` operation. The current implementation looks like the `nextLetterCandidate$` needs some value from the `switchMap` but it doesn't. This should be also faster, because you don't have to wait for the `this.audioService.getAudioBuffer` to complete before returning the `candidate`? (I didn't test the proposal, read it as a pseudo code)

aorin: Looks good but I think it's better if these operations are in the same subscription so that they can be easily unsubscribed. So I removed the _getAudioBuffer_ call from the _nextLetterCandidate_ observable and moved it to _nextCandidateToCacheSub_. Now the names should be more descriptive.

olzraiti: Ok! I don't see those changes yet, did you maybe forget to push?

aorin: commits should be visible below ([this](https://github.com/luomus/laji/pull/80/commits/8f4aba75c50355b5465422b303191192764705c8) is the commit)

olzraiti (approved):

PR #81, opened 2021-03-24
Kerttu letter statistics 177380329

aorin:
Testable [here](https://177380329.dev.laji.fi/theme/kerttu/result) (but you have to be logged in as a user that have many letter annotations before seeing any numbers).

Adds a table to kerttu result page that contains letter statistics for the user grouped by species. Also shows mean similarity value (how similar user's annotations are compared to other users).


projects/laji/src/app/+theme/kerttu/kerttu-result/kerttu-letter-result-table/kerttu-letter-result-table.component.ts
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    this.loading = true;
+    this.nameSub = this.triplestoreLabelService.getAll(ids, this.translate.currentLang).subscribe(res => {
+      this._data.forEach(d => { d.species = res[d.taxonId]; });
+      this._data = [...this._data];
+      this.loading = false;
+      this.cd.markForCheck();
+    });
+  }
```
olzraiti, line 93: Maybe using a cell template for the species label would be more elegant? (no need for creating pipe, subscribing/unsubscribing, resetting _data & touching the change detector...). I couldn't find an existing pipe for this by taking a quick glance, but an example solution is the winter bird species result service (`projects/laji/src/app/+project-form/results/wbc-result/wbc-species/wbc-species-list/wbc-species-list.component.ts`)

olzraiti (changes requested):

projects/laji/src/app/+theme/kerttu/kerttu-result/kerttu-letter-result-table/kerttu-letter-result-table.component.ts
aorin: There already exists a label pipe but the problem is that the automatic sorting doesn't work then. Some kind of custom sorting would need to be implemented (like in the species list `projects/laji/src/app/+taxonomy/species/species-list/species-list.component.ts`). This is probably quite a common problem so it should be on the laji-datatable component. But it sounds like a quite big task since it would possibly affect other tables so maybe it should be its own task?

olzraiti: Ok!

olzraiti (approved):

PR #82, opened 2021-03-24
LajiForm receives the profile metadata setttings with default value [#177316308]

olzraiti:
Currently the metadata settings default value doesn't work, because the defaults were locally in profile component only. I moved getting the profile to user service, so that it can be properly prepared there.


vixriihi (approved):

vixriihi:
LGTM

PR #83, opened 2021-03-24
Taxon identification #174428349 #175142664 #177454917

Blodir:
https://175142664.dev.laji.fi/taxon/MX.1/identification


projects/laji/src/app/+taxonomy/taxon/info-card/taxon-identification/taxon-identification.component.ts
```diff
@@ -1,28 +1,116 @@
...
+        this.loading = false;
+        this.cdr.markForCheck();
+        this.totalChildren$.pipe(take(1)).subscribe(total => {
+          if (this.children.length < total) {
+            setTimeout(() => this.triggerInfiniteScrollStatusCheck.next(), 0);
+          }
+        });
+      })
+    );
+  }
```
olzraiti, line 87: This part was a bit hard to understand for me, since the two subscriptions can't be read as a stream. What do you think about the following:

1. Make `loadChildDataSource` to return just an observable of the children result (`PagedResult<Taxonomy>`)
2. Remove `taxonChange$`, instead use `set` for  `@Input() taxon`, so that the taxon value would come from a single observable
3. Init `this.subscription` like this:

```
    this.subscription = this.taxon$.pipe(
      tap(() => {
        this.children = [];
        this.loading = true;
        this.cdr.markForCheck();
      }),
      switchMap(taxon => this.facade.loadChildDataSource(taxon)),
      tap(result => {
        this.children = result.children;
        this.loading = false;
        this.cdr.markForCheck();
        if (this.children.length < result.total) {
          this.triggerInfiniteScrollStatusCheck.next();
        }
      })
    ).subscribe();
```

You could also get rid of the subscription and possibly the change detection checks by removing the subscription and instead map the pipe value to some view model observable and use it with the `async` pipe.

This proposal is pretty much a different opinionated approach I tend to use to get rid of subscription usage to avoid  making the code less stateful (less subscriptions -> less state). Your solution seems to work so it's ok as it is so no need for changes if you want to keep the current implementation (also, I didn't test this proposal).

projects/laji/src/app/+taxonomy/taxon/info-card/taxon-identification/taxon-identification.component.ts
```diff
@@ -1,28 +1,116 @@
...
+        rect.bottom <= (this.window.innerHeight || this.document.documentElement.clientHeight) &&
+        rect.right <= (this.window.innerWidth || this.document.documentElement.clientWidth)
+    );
+  }
+
+  isWithinXPixelsOfViewport(element: Element, px: number) {
+    const rect = element.getBoundingClientRect();
+    return (
+      this.window.innerHeight - rect.y > -1 * px
+      || this.document.documentElement.clientHeight - rect.y > -1 * px
```
olzraiti, line 113: A minor thing, but instead of `-1 * px` how about just `-px`?

olzraiti (changes requested):

olzraiti:
Conflicts need to be fixed

projects/laji/src/app/+taxonomy/taxon/info-card/taxon-identification/taxon-identification.component.ts
Blodir: * Unfortunately async pipe can't be used here, because `children$` needs to be subscribed to after `ViewInit`
* `taxonChange$` also exists, because the subscription has to be inside `ViewInit` (neither `OnChanges` or a setter can be directly used here)
* `loadChildDataSource` returns `void` because of the [facade pattern](https://medium.com/angular-in-depth/angular-you-may-not-need-ngrx-e80546cc56ee). It seems to be complicating things here though 🤔 
* I suppose the two subscriptions @ `TaxonIdentificationComponent` could technically be joined with a `switchMap`, but I think it is more logical to keep them separated as there's no reason to resub to `children$` every time a taxonChange$ is emitted.
* note about setTimeout since your proposal removed it: it's very important that `triggerInfiniteScrollStatusCheck.next()` is called asynchronously, because otherwise we'd end up in an infinite loop. The way this works is that whenever `infiniteScrollStatusCheck` emits we check if `#loadMore` is within INFINITE_SCROLL_DISTANCE of bottom of the viewport, if it is, then we attempt to load more children. This loop continues until `#loadMore` is no longer within distance or all children have been loaded. The `setTimeout` yields execution rights to the renderer, so that the `#loadMore` elements position in the DOM is updated before we check again.

olzraiti: Right! Thanks for explaining the reasoning :+1:

olzraiti (approved):

PR #84, opened 2021-03-24
Update laji-form to style framework agnostic version

olzraiti:
Demo: https://177458001.dev.laji.fi/

The latest `laji-form` is style framework agnostic, and the theme must be passed by the client.

This update results in quite many build warnings in like this:

```
Warning: /home/local/olzraiti/src/luomus/laji/node_modules/laji-form/lib/themes/bs3.js depends on 'react-bootstrap/lib/Col'. CommonJS or AMD dependencies can cause optimization bailouts.
For more info see: https://angular.io/guide/build#configuring-commonjs-dependencies
```

This is because I imported the `react-bootstrap` components like this:
```
import * as Row from 'react-bootstrap/lib/Row';
import * as Col from 'react-bootstrap/lib/Col';
...
```

instead of like this:
```
import { Row, Col, ...} from 'react-bootstrap';
```

since it should result in a smaller build since everything from `react-bootstrap` isn't imported?

 I'm not sure if these warnings create any problems other than cluttering build log. There are other identical warnings from other code than `laji-form`.

EDIT: Apparently the warning's aren't show in the feature branch demo.


olzraiti:
Added also laji-form typescript declaration files.

projects/laji/src/app/shared-modules/laji-form/laji-form/laji-form.component.ts
```diff
@@ -24,6 +24,8 @@ import { TranslateService } from '@ngx-translate/core';
 import { PersonApi } from '../../../shared/api/PersonApi';
 import { combineLatest } from 'rxjs';
 import { Profile } from '../../../shared/model/Profile';
+import LajiForm from 'laji-form/lib/index';
+import { Theme as LajiFormTheme } from 'laji-form/lib/themes/theme';
```
olzraiti, line 28: I don't remember why we import laji-form asynschronously, but I wonder if importing these here without async breaks something that the async laji-form import is fixing?

vixriihi: That was added so that the bundle size for vihko/project main pages could be kept as small as possible. You can run `npm run bundle-report` to get the report to see what's in what bundle (you might have to change dist/browser/stats-es2015.json to dist/browser/stats.json that neme seems to be changing quite often 😂),

I ran those in this branch and all looks good 👍

vixriihi (approved):
LGTM

projects/laji/src/app/shared-modules/laji-form/laji-form/laji-form.component.ts
```diff
@@ -124,28 +125,29 @@ export class LajiFormComponent implements OnDestroy, OnChanges, AfterViewInit {
...
+    combineLatest(
+      import('laji-form'),
+      import('laji-form/lib/themes/bs3'),
+      this.userService.getUserSetting<any>(this.settingsKey).pipe(
+        concatMap(settings => this.userService.getUserSetting<any>(GLOBAL_SETTINGS).pipe(
+          map(globalSettings => ({...globalSettings, ...settings}))
+        )),
+        take(1)
+      ),
+      this.userService.getProfile().pipe(map(profile => profile.settings?.defaultMediaMetadata))
```
vixriihi, line 144: Since there was no take(1) in this I added one to getProfile in UserService (since that looked like a method that should only return one result) otherwise this subscription would not complete (not an issue in here, but could become issue somewhere else)

vixriihi (approved):

PR #85, opened 2021-03-25
Import fails if end date not given [fixes #177488749]

vixriihi:
fix undefined value becoming `undefined` string causing error in the date validator


projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -338,7 +338,7 @@ export class MappingService {
       return value.toISOString();
     }
 
-    return String(value);
+    return String(value || '');
```
olzraiti, line 341: This would replace `"0"` with `""`. How about `value ?? ''` ?

olzraiti (changes requested):

projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
olzraiti: I didn't read the whole code context. In this problem domain, the value shouldn't be `"0"` ever anyway.

olzraiti (approved):

projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
vixriihi: nw :)

PR #89, opened 2021-04-07
Red list & sensitive status visualization 175455761

MLumme:
Added new column to observation list for species sensitivity, and modified redlist column to show coloured box with the status.


wkmor1:
Would it be possible to make the Red list labels middle rather than bottom aligned within the table rows. At the moment they are they are the only column that look like this so looks a bit odd. Otherwise looks good to me.

MLumme:
> Would it be possible to make the Red list labels middle rather than bottom aligned within the table rows. At the moment they are they are the only column that look like this so looks a bit odd. Otherwise looks good to me.

Sure, I'll see what can be done.

MLumme:
The icon is now centered, but only way I managed to do it was using relative positioning, so if the datatable cell height changes at some point it will require fixing.

wkmor1:
Cool. Will it be a problem if font-size changes? @Blodir is working on accessibility issues on the site. Not sure if our default fonts might be getting bigger soon?

MLumme:
> Cool. Will it be a problem if font-size changes? @Blodir is working on accessibility issues on the site. Not sure if our default fonts might be getting bigger soon?

Well, depends on how/if the datatable cell changes size with the font, as my current problem is that the box creating the icon is hardcoded to be 25px high, which actually higher than the area the cell reserved for content, which caused it to be pushed down.

But, as it is now, and expecting that the cell won't grow, the text would grow inside the box but would stay centered, and escape it once the font is over 25px in height. If the cell changes size, then the box would be displaced, but the text should still be centered in its box, offset as it is and possibly grown out of it.

And just to be clear, there might still be a way to force it centered some other way, as I'm not exactly that experienced with CSS.

wkmor1:
Yeah right. My CSS chops are probs even less hot. Maybe @olzraiti or @vixriihi have an idea?

olzraiti:
I pushed a fix. If we want to get automatic vertical centering in this case, we need to adjust how `datatable-body-cell` (grandparent of the iucn cell template) aligns it's content (in this case centers it vertically), because the cell template doesn't have any idea about the cell height. I was a bit worried that this might break some other cell templates aligning, since this affects all body cells, but I tested the taxon & observation datatables with all columns and it works fine. I didn't find any cell template styles in the source code that wouldn't be used in those tables.

MLumme:
@olzraiti seems that changing the datatable cell to a use flex display broke the horizontal centering of the sensitivity- and IUCN-icons somehow.

olzraiti:
That seems to be the case for Firefox indeed, on Chrome it's working properly... This is why I was worried about using flexbox :) I'll look into tomorrow.

olzraiti:
Fixed by adding a new column class to center content horizontally, which I added to the iucn & sensitivity columns.

MLumme:
Is this now in such a state that it could be finally merged to dev?

olzraiti (approved):
Yes, LGTM

PR #90, opened 2021-04-13
Vihko button styling [#177593782]

olzraiti:
Makes the Vihko buttons not use primary variant by default. Demo: https://177593782.dev.laji.fi/project/JX.519

Implements also styling for bootstrap active button & fixes button group (both fixes can be seen at the images grid size button: 

Broken here:  https://beta.laji.fi/taxon/browse
Fixed here: https://177593782.dev.laji.fi/taxon/browse


vixriihi (approved):
LGTM

PR #91, opened 2021-04-13
fix bug in taxon card content tab when switch to other specie - 177659751

Hvar83:
The problem is in the taxon-card and it happens when i move between tabs and later i switch to another specie.

For example:
- go to taxon MX.44058
- move between tabs (at least open "occurrences" and "specimens")
- switch to other species by using dropdown menu in breadcrumbs. 
- if the next specie i open has less tabs than the previous, for example MX.44052, the content is not correct. 

2 components are opened at same time. At least i could see "specimens" and "occurrences" or "specimens" and "taxonomy" opened at same time.

### DIFF
--- a/projects/laji/src/app/+taxonomy/taxon/info-card/info-card.component.ts
+++ b/projects/laji/src/app/+taxonomy/taxon/info-card/info-card.component.ts
@@ -148,7 +148,6 @@ export class InfoCardComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   updateRoute(id: string, tab = this.activeTab, context = this.context, replaceUrl = false) {
-    this.loadedTabs.reset();
     this.activeTab = tab;
     this.routeUpdate.emit({id: id, tab: tab, context: context, replaceUrl: replaceUrl});
     this.cd.markForCheck();
@@ -217,7 +216,6 @@ export class InfoCardComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   private getIsEndangered(taxon: Taxonomy): boolean {
-    this.cd.detectChanges();
     if (!taxon.latestRedListStatusFinland) {
       return false;
     }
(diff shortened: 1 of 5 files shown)


projects/laji/src/app/+taxonomy/taxon/info-card/info-card.component.ts
```diff
@@ -148,8 +148,12 @@ export class InfoCardComponent implements OnInit, OnChanges, OnDestroy {
   }
 
   updateRoute(id: string, tab = this.activeTab, context = this.context, replaceUrl = false) {
-    this.selectedTab = tab;
+    if (this.activeTab !== 'endangerment' && this.activeTab !== 'invasive') {
+      this.loadedTabs.reset();
+    }
```
vixriihi, line 153: changing tab causes updateRoute so resetting loadedTabs on almost all tabs is not ideal. (idea with loaded tabs is to keep data for one specific taxon always in memory and not to re-fetch the same data on every tab change)

vixriihi (changes requested):

vixriihi (approved):
LGTM

vixriihi (changes requested):
not ready yet

esko-piirainen:
There is a bug with this pull request found in beta - details in pivotal

Blodir:
I have now fixed the bug that was mentioned in pivotal:

> https://177659751.dev.laji.fi/taxon/MX.66113/endangerment
> when you switch between taxa (using species dropdown in breadcrumb) and then back to the species the contents do not load

This was a separate, but somewhat related bug. Took me a very long time to figure it out despite the fix being only 2 lines long. I'll describe the issue here as I understand it.

Tabs component uses content projection to populate the content of each tab. I was falsely under the assumption that the change detector hierarchy would be linear ie.
```
<parent>
  <tabs>
    <tab>
      <content></content>
    </tab>
  </tabs>
</parent>
```
This is what the markup suggests. However content projection actually breaks the hierarchy and in fact `tabs`, `tab` and `content` are all direct children of `parent`. In order to run changedetection for `content` we actually have to call `detectChanges()` in `parent`.

After this fix everything seems to work fine.

Blodir:
https://177659751.dev.laji.fi/

esko-piirainen (approved):
ok

PR #92, opened 2021-04-13
Fix SSR error on theme pages

aorin:
Fixes a problem that causes theme pages (https://beta.laji.fi/theme) load very slowly.


projects/laji/src/app/+theme/kerttu/audio-viewer/audio-info/audio-info-map/audio-info-map.component.ts
```diff
@@ -1,8 +1,7 @@
 import { Component, ChangeDetectionStrategy, ViewChild, Input } from '@angular/core';
 import { LajiMapComponent } from '@laji-map/laji-map.component';
-import { LajiMapOptions } from '@laji-map/laji-map.interface';
+import { LajiMapOptions, LajiMapTileLayerName, LajiMapDataOptions } from '@laji-map/laji-map.interface';
```
vixriihi, line 3: in version 3.8 of typescript we got [type only imports (and exports)](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-8.html) that could have been used to fix this also, but this is fine also

vixriihi (approved):
LGTM

PR #93, opened 2021-04-14
fix issue on document annotation images - [#177723306]

Hvar83:
In the document-annotation document, the image component didn't change the main picture. 
User move to next or previous by shortcuts, but when the observation is loaded, the main picture is always the same.
That happens when the next or the previous are from the same observation of the same user


vixriihi (approved):
lgtm

PR #97, opened 2021-04-15
fix kerttu buttons style

aorin:
Testable [here](https://177776349.dev.laji.fi/theme/kerttu/letters). Fixes these issues with Kerttu buttons:
- "I don't know" button is smaller than the others
- Time and place text is not aligned with audio viewer buttons
- Active audio viewer buttons look weird


vixriihi (approved):
LGTM

PR #99, opened 2021-04-16
refine button active state styles

Blodir:
subtle global improvements to button styles


Blodir:
https://123456.dev.laji.fi/

PR #102, opened 2021-04-19
Auto opening selects are hiding others

Hvar83:
The problem is happening inside metadata-select. When user click some checkbox, if a checkbox with filter has some option checked, it will open automatically

### DIFF
--- a/projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
+++ b/projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
@@ -16,6 +16,7 @@ import { AnnotationService } from '../../document-viewer/service/annotation.serv
 import { MultiLangService } from '../../lang/service/multi-lang.service';
 import { Annotation } from '../../../shared/model/Annotation';
 import { SelectOptions } from '../select-subcategories/select-subcategories.component';
+import { FilterOptionsPipe } from '../../search-filters/filter-options.pipe';
 
 export enum SelectStyle {
   basic,
@@ -129,13 +130,13 @@ export class MetadataSelectComponent implements OnChanges, OnDestroy, ControlVal
 
     this.subOptions = (this.options ? byOptions$ : byField$).pipe(
       switchMap(options => this.mapToWarehouse ? this.optionsToWarehouseID(options) : of(options)),
+      map(options => this.deleteOptionsfromExcludedOptions(options, this.excludeOptions)),
       map(options => this.labelAsValue ? options.map(o => ({...o, id: o.value})) : options),
       map(options => this.firstOptions?.length > 0 ? this.sortOptionsByAnotherList(options) : (
         this._shouldSort ? options.sort((a, b) => a.value.localeCompare(b.value)) : options
       ))
     ).subscribe(options => {
         this.setOptions(options);
-        this._options = this.deleteOptionsfromExcludedOptions(this._options, this.excludeOptions);
         this.initActive();
         this.cd.markForCheck();
       });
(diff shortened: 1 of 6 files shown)


projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
```diff
@@ -134,6 +135,7 @@ export class MetadataSelectComponent implements OnChanges, OnDestroy, ControlVal
       ))
     ).subscribe(options => {
         this.setOptions(options);
+        this._options = this.deleteOptionsfromExcludedOptions(this._options, this.excludeOptions);
```
vixriihi, line 138: this could be moved to the pipe so that options here would be already correct. For example just above line 132 (then using `labelAsValue` wouldn't break this behavior). Somethinkg like:
 ```typescript
 map(options => this.deleteOptionsfromExcludedOptions(options, this.excludeOptions))

projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
```diff
@@ -165,6 +167,22 @@ export class MetadataSelectComponent implements OnChanges, OnDestroy, ControlVal
...
+
+    options = options.reduce((filtered, current) => {
+      const id = current['id'].replace('MX.', '');
+      if (excludeOptions.indexOf(id) === -1) {
+        filtered.push(current);
+      }
+      return filtered;
+    }, []);
+
+    return options;
```
vixriihi, line 183: The replace 'MX' refers to taxon object and in the context of metadata select there shouldn't be anything about it.  Ideally you could just use simple 
```typescript
return options.filter(options => !excludeOptions.includes(option.id))`
```

projects/laji/src/app/shared-modules/select/select/select.component.ts
```diff
@@ -65,7 +66,9 @@ export class SelectComponent<T extends idType|SelectOption = string> implements
       this.selected = [];
       this.open = false;
     }
-    this.initOptions(this.selected);
+    if (!this.selected || !this.selected.length || (this.selected.length !== this.selectedOptions.length)) {
+      this.initOptions(this.selected);
+    }
```
vixriihi, line 71: capture what has changed on this `ngOnChanges` method (1st parameter in that). and then use that to decide if the `this.initOptions` should be called. (looks like it should be called at least when `options`, `checkboxType`, `disabled` and `selected` inputs change)

vixriihi (changes requested):

projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
Hvar83: Agree

projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
Hvar83: @vixriihi ExcludedOptions are without MX. prefix, so options will be the same after filtering. I moved the change to upper component species-form, but i don't think my solution is good.

projects/laji/src/app/shared-modules/select/select/select.component.ts
Hvar83: @vixriihi i think i verified all conditions, but i don't understand if the condition has to be better presented

projects/laji/src/app/shared-modules/select/metadata-select/metadata-select.component.ts
vixriihi: there seems to be at least 3 places now where the MX is added to the value.  In the `species-form.compoent.ts` it might be better to just rename all invasiveStatuses as they really are (since that real key is already visible in the query parameter)

vixriihi (approved):

PR #105, opened 2021-04-30
Datatable sorting refactoring 177743192

aorin:
Testable [here](https://177743192.dev.laji.fi).

Currently sorting works only on [taxon table](https://177743192.dev.laji.fi/taxon/list?informalGroupFilters=MVL.27&onlyFinnish=true) when using client side sorting and async column templates such as the label template. I refactored sorting implemented there to the general laji-datatable component. I also refactored [kerttu result tables](https://177743192.dev.laji.fi/theme/kerttu/result) by deleting code related to sorting.


projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -273,4 +309,89 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
...
+          this.sortLoading = false;
+          this.changeDetectorRef.markForCheck();
+        }
+      );
+    } else {
+      this.sortLoading = false;
+    }
+  }
+
+  private sort(sorts: DatatableSort[], rows: any[]): Observable<any> {
```
olzraiti, line 332: Return type should be `Observable<any[]>`

projects/laji/src/app/shared-modules/datatable/model/datatable-column.ts
```diff
@@ -5,4 +5,10 @@ export interface DatatableColumn extends TableColumn {
   target?: string;
   info?: string;
   required?: boolean;
+  sortTemplate?: string;
+}
+
+export interface DatatableSort {
+  prop: string;
+  dir: 'asc'|'desc';
```
olzraiti, line 13: `dir: 'asc' | 'desc';` (add spaces)

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -273,4 +309,89 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
...
+      .subscribe(() => {
+          this._rows = [...this._rows];
+          this.sortLoading = false;
+          this.changeDetectorRef.markForCheck();
+        }
+      );
+    } else {
+      this.sortLoading = false;
+    }
+  }
```
olzraiti, line 330: The code is a bit had to follow since it's doing the sorting in a mutable way. Looking at this:
```
      this.sortSub = this.sort(sorts, this._rows)
      .subscribe(() => {
          this._rows = [...this._rows];
```
It looks like a pipe without any return value is called, and you kind of have to guess that the pipe sorted the `this._rows` before it's reassigned to a new variable.

Doing mutable operations is a common source of confusion and therefore bugs, and since we are anyway reassigning the mutably sorted `this._rows`, I think this would be more readable if the operations were done in an immutable way.

The rest of the comments are guidelines for how to get the sorting done immutably.

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -273,4 +309,89 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
...
+        }
+      });
+      return arr;
+    }, []);
+
+    return (obs.length > 0 ? forkJoin(obs) : of([]));
+  }
+
+  private customSort(sorts: DatatableSort[], results: any[]) {
+    return results.sort((a: any, b: any) => {
```
olzraiti, line 374: Make this immutable and make it clear in the signature:

1. Add return type `any[]`
2. `return results.sort(...` -> `return [...results].sort(...`

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -273,4 +309,89 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
...
+    });
+  }
+
+  private defaultSort(results: any[]) {
+    const values = [...results];
+
+    values.forEach((val) => {
+      results[val.preSortIndex] = val;
+    });
+  }
```
olzraiti, line 396: Make this immutable and make it clear in the signature:

1. Add return type `any[]`
2. Update algorithm to:
``` 
return [...results].map((val) => {
      results[val.preSortIndex] = val;
    });
```

olzraiti (changes requested):
Always happy to see changes that remove code more than they add :)

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -273,4 +309,89 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
...
+        }
+      );
+    } else {
+      this.sortLoading = false;
+    }
+  }
+
+  private sort(sorts: DatatableSort[], rows: any[]): Observable<any> {
+    return this.setSortValues(sorts, rows)
+      .pipe(map(() => {
```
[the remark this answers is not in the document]
aorin: I thought about removing the `this.sortValues` variable but it is used as a cache so I left it as it was. It seemed like a good idea to fetch observables only once when changing sorting.

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -63,7 +66,14 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnDestroy {
   @Input() selected: any[] = [];
 
   initialized = false;
+  sortLoading = false;
   private filterChange$ = new Subject();
+
+  private prevSort: DatatableSort[] = [];
```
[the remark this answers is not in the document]
aorin: yes, it is just the current sort, changed it to `sorts`

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
aorin: @olzraiti Good suggestion, changes related to this should be now done

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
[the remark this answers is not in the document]
olzraiti: Fair enough, performance wise it's a good solution

olzraiti (approved):
:+1:

PR #107, opened 2021-05-04
Allow ETRS-TM35FIN grid coordinates on excel-import [#176397033]

vixriihi:
This allows ETRS-TM35FIN grid coordinates on excel-import

testable at https://176397033.dev.laji.fi

### DIFF
--- a/projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
+++ b/projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
@@ -1,7 +1,7 @@
 import { Injectable } from '@angular/core';
 import { TranslateService } from '@ngx-translate/core';
 import { IColMap, IFormField, IUserMappings, IValueMap, TUserValueMap, VALUE_IGNORE } from '../model/excel';
-import { convertAnyToWGS84GeoJSON } from 'laji-map/lib/utils';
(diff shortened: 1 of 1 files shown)


projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
```diff
@@ -428,7 +428,10 @@ export class MappingService {
         const ykjParts = value.split(':');
         if (ykjParts[0].length === ykjParts[1].length) {
           try {
-            return this.coordinateService.convertYkjToGeoJsonFeature(ykjParts[0], ykjParts[1]).geometry;
+            const data = latLngGridToGeoJSON([ykjParts[0], ykjParts[1]]);
+            if (data.geometry) {
+              value = data.geometry;
+            }
```
olzraiti, line 434: Should we just return the `data.geometry` here instead of binding it to the `value` variable? Now the execution will continue to trying to `convertAnyToWGS84GeoJSON` (13 lines below), and if that fails, `null` will be returned. There could be cases when data that is valid for `latLngGridToGeoJSON` but invalid for `convertANyToWGS84GeoJSON`, and the value from `latLngGridToGeoJSON` would be lost.

olzraiti (changes requested):

olzraiti (approved):
:+1:

PR #109, opened 2021-05-04
Protax submit fails when submitting a modified file

aorin:
Testable [here](https://177202324.dev.laji.fi/theme/protax). A file input is cleared after submitting it to reduce situations where user submits a modified file which causes problems on some browsers.

### DIFF
--- a/projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
+++ b/projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
@@ -1,6 +1,6 @@
-import {Component, ChangeDetectionStrategy, Output, EventEmitter, Input} from '@angular/core';
-import {ProtaxModelEnum} from '../models';
-import {DialogService} from '../../../shared/service/dialog.service';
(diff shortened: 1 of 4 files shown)


projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
```diff
@@ -14,7 +14,9 @@ enum Tab {
   styleUrls: ['./protax-form.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class ProtaxFormComponent {
+export class ProtaxFormComponent implements OnChanges {
+  @ViewChild('fileInput') fileInput: ElementRef;
```
vixriihi, line 18: this could be typed to `ElementRef<HTMLInputElement>` so that the value used bellow wouldn't have to come from any

vixriihi (approved):

projects/laji/src/app/+theme/protax/protax-form/protax-form.component.ts
aorin: type added

PR #110, opened 2021-05-04
Don't show info page placeholder if no content

aorin:
Testable on this page: https://177961515.dev.laji.fi/project/MHL.69/about. Previously info page showed a loading placeholder also when it had loaded the content but the content was empty.


vixriihi (approved):
LGTM

PR #111, opened 2021-05-04
Collection tree filter 2 175492731

MLumme:


projects/laji/src/app/shared-modules/select-collections/pipe/collection-url.pipe.ts
```diff
@@ -0,0 +1,13 @@
...
+@Pipe({
+  name: 'collectionUrl'
+})
+export class CollectionUrlPipe implements PipeTransform {
+
+  transform(value: any): any {
+    return IdService.getUri(value);
+  }
+
+}
```
vixriihi, line 13: This is almost exactly the same as `toFullUri` pipe. You can remove this and use that instead in all places where you're using `collectionUrl` (output is exactly the same).

vixriihi (changes requested):
small tweak

projects/laji/src/app/shared-modules/select-collections/pipe/collection-url.pipe.ts
MLumme: Ok, will do.

MLumme: Fixed now.

vixriihi (approved):

PR #112, opened 2021-05-05
Fix filters after navigating from named place edit view [#178035225]

olzraiti:
Steps to repro:

1. Make sure that you are admin of water bird form
2. Goto https://beta.laji.fi/project/MHL.65/form/MHL.65/places
2. Select some place
3. Click named place edit button of the selected place
4. Click cancel edit on the footer

### What happens?
Query contains filters for municipality, which can't be edited in the UI
### What should happen?
Query doesn't contain filters that aren't editable in the UI


aorin (approved):
Looks good

PR #114, opened 2021-05-06
Various keyboard navigability improvements 177593449

Blodir:
https://www.pivotaltracker.com/story/show/177593449

https://177593449.dev.laji.fi/


olzraiti (changes requested):
I think the `luA11yClick` should be replaced with ARIA-roles for several reasons:

1. ARIA roles provide important information for the browser about the element (e.g. screen readers, also if I remember correct you also use the excellent Vimium chrome plugin: Vimium will detect ARIA roles, so it displays them as clickable. Vimium won't detect the `luA11y`)
2. It's vanilla, any front end developer know what they mean, on the other hand `luA11yClick` is pretty cryptic (some might know what a11y is, but what does "a11y click" mean?)
3. No JS required

Most of the `luA11yClick` could be replaced with `role="button"`.

Unfortunately the `lu-button` already uses `role` attribute for styling. How about renaming the `lu-button` `role` to `variant`? Then it would use the same term as laji-form (which inherited the term from bootstrap 4).

`luA11yClose` can't be replaced with ARIA AFAIK, but it's pretty cryptic. How about `closeOnEsc`?

projects/laji-ui/src/lib/checkbox/checkbox.component.scss
```diff
@@ -29,14 +29,14 @@ input {
...
   &:checked::after {
     position: absolute;
     left: 0.15em;
     top: -0.15em;
     content: "✓";
     color: $neutral-7;
   }
+  &:focus:not(:focus-visible) {
+    outline: none;
+  }
```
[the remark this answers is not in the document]
vixriihi: Even `:not` is not supported by IE11 :)

The decision to drop IE 11 support was made on Friday so these shouldn't be problem anymore, but instead of having each individual component to decide what to do with focus outline I think this should be made a global rule.

Blodir:
> I think the `luA11yClick` should be replaced with ARIA-roles for several reasons:
> 
> 1. ARIA roles provide important information for the browser about the element (e.g. screen readers, also if I remember correct you also use the excellent Vimium chrome plugin: Vimium will detect ARIA roles, so it displays them as clickable. Vimium won't detect the `luA11y`)
> 2. It's vanilla, any front end developer know what they mean, on the other hand `luA11yClick` is pretty cryptic (some might know what a11y is, but what does "a11y click" mean?)
> 3. No JS required
> 
> Most of the `luA11yClick` could be replaced with `role="button"`.
> 
> Unfortunately the `lu-button` already uses `role` attribute for styling. How about renaming the `lu-button` `role` to `variant`? Then it would use the same term as laji-form (which inherited the term from bootstrap 4).
> 
> `luA11yClose` can't be replaced with ARIA AFAIK, but it's pretty cryptic. How about renaming it `closeOnEsc`?

[a11yClick can't be replaced with role="button" as it doesn't add a keylistener](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role#accessibility_concerns). However adding a role="button" in addition is a good idea :)

luA11yClose -> luCloseOnEsc sounds good, similarly could do luA11yClick -> luClickOnEnter.

`lu-button` should be removed entirely and replaced with classes.

projects/laji-ui/src/lib/directives/a11y-click.directive.ts
```diff
@@ -0,0 +1,19 @@
...
+
+@Directive({
+  selector: '[luA11yClick]'
+})
+export class a11yClickDirective {
+  constructor(private el: ElementRef) {}
+
+  @HostListener('window:keyup', ['$event'])
+  keyEvent(event: KeyboardEvent) {
+    if (event.target === this.el.nativeElement && event.keyCode === KEY_CODE.ENTER) {
```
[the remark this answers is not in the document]
Blodir: There was a small error here that caused the performance issues: the event handler was bound to "window" rather than the host element. The issue is now fixed :)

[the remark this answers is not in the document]
vixriihi: ok that helps with the callbacks yes but this would still register 176 individual event listeners on the observation page. I know on older browsers that causes performance issues, but since we dropped IE11, this is okay.

Blodir:
I added the `role="button"` to affected elements, but removing `lu-button` has to go to a different task/branch. This PR should be ready now.

projects/laji-ui/src/lib/checkbox/checkbox.component.html
```diff
@@ -1 +1 @@
-<label [ngClass]="{'checked': isChecked, 'disabled': disabled}"><input [disabled]="disabled" type="checkbox" (input)="onInput($event)" luClickOnEnter #checkbox><ng-content></ng-content></label>
+<label [ngClass]="{'checked': isChecked, 'disabled': disabled}"><input [disabled]="disabled" type="checkbox" role="checkbox" (input)="onInput($event)" luClickOnEnter #checkbox><ng-content></ng-content></label>
```
vixriihi, line 1: Adding `role="checkbox"` to actual checkbox is not needed since the input type checkbox is already semantically correct. Role is only needed if there is an element that is behaving like a checkbox, but that is not an actual checkbox input element. Also when using `role="checkbox"` the `aria-checked` attribute should also be present

vixriihi (changes requested):

olzraiti:
> [a11yClick can't be replaced with role="button" as it doesn't add a keylistener](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role). However adding a role="button" in addition is a good idea :) Btw vimium should detect the `tabindex="0"` even without the `role="button"`.

Thanks for the link, I tested again and you're right :) I tested this when I did the PR but apparently I didn't do it properly.

From reading that article it seems that we should also make space-key trigger a click. Button elements are clickable via space, so it would make sense. Could we include that in this PR? `luClickOnEnterAndSpace` sounds a bit awkward. How about `keyboardClickable`? I'm not sure if `lu` prefix is needed, it sounds to me like something that would work only for lu-prefixed elements. On the other hand, with the prefix other devs will understand right away that it's our custom directive and that's good, so idk :)

projects/laji/src/app/+theme/theme.component.html
```diff
@@ -8,127 +8,127 @@ <h1 class="mt-6">
...
-    <a [routerLink]="['/theme/protax'] | localize" *lajiForTypes="['dev']" tabindex="-1" class="navbox">
+    <a [routerLink]="['/theme/protax'] | localize" *lajiForTypes="['dev']" tabindex="0" class="navbox">
       <h5>{{ 'theme.protax' | translate }}</h5>
       <p>{{ 'theme.protax.intro' | translate }}</p>
     </a>
-    <a [routerLink]="['/theme/publications'] | localize" tabindex="-1" class="navbox">
+    <a [routerLink]="['/theme/publications'] | localize" tabindex="0" class="navbox">
       <h5>{{ 'finbif-bib.title' | translate }}</h5>
       <p>{{ 'finbif-bib.intro' | translate }}</p>
     </a>
```
olzraiti, line 134: `a` is tab accessible even without tabindex. This doesn't matter much since having it here doesn't affect the elements in any way, but  it's redundant.

Blodir: fixd

projects/laji-ui/src/lib/checkbox/checkbox.component.html
Blodir: fixd

Blodir:
> > [a11yClick can't be replaced with role="button" as it doesn't add a keylistener](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/button_role). However adding a role="button" in addition is a good idea :) Btw vimium should detect the `tabindex="0"` even without the `role="button"`.
> 
> Thanks for the link, I tested again and you're right :) I tested this when I did the PR but apparently I didn't do it properly.
> 
> From reading that article it seems that we should also make space-key trigger a click. Button elements are clickable via space, so it would make sense. Could we include that in this PR? `luClickOnEnterAndSpace` sounds a bit awkward. How about `keyboardClickable`? I'm not sure if `lu` prefix is needed, it sounds to me like something that would work only for lu-prefixed elements. On the other hand, with the prefix other devs will understand right away that it's our custom directive and that's good, so idk :)

Good point. Links however aren't activated with space by default 🤔 I will fix it though, don't see any harm in adding space.

I think namespacing the directives (ie. lu-prefix) makes sense to avoid possible overlap with other libraries etc.

olzraiti:
Lookin good! After the conflicts are fixed this is ready.

olzraiti (approved):
:+1:

PR #115, opened 2021-05-06
Change some vihko buttons to links 177889264

Blodir:
https://www.pivotaltracker.com/story/show/177889264

https://177889264.dev.laji.fi/

also fixes: https://www.pivotaltracker.com/story/show/177593502 (seems to be duplicate task)

### DIFF
--- a/projects/laji/src/app/shared-modules/latest-documents/latest/short-document.component.ts
+++ b/projects/laji/src/app/shared-modules/latest-documents/latest/short-document.component.ts
@@ -89,10 +89,8 @@ export class ShortDocumentComponent implements OnInit, OnChanges, OnDestroy {
     return getLocality$(this.translate, this.labelService, gatheringInfo, this.document);
   }
 
-  editDocument(formId, documentId) {
-    this.router.navigate(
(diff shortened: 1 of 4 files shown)


projects/laji/src/app/shared-modules/latest-documents/latest/short-document.component.ts
```diff
@@ -89,10 +89,8 @@ export class ShortDocumentComponent implements OnInit, OnChanges, OnDestroy {
     return getLocality$(this.translate, this.labelService, gatheringInfo, this.document);
   }
 
-  editDocument(formId, documentId) {
-    this.router.navigate(
-      this.localizeRouterService.translateRoute([this.formService.getEditUrlPath(formId, documentId)])
-    );
+  getEditDocumentRoute(formId, documentId) {
```
vixriihi, line 92: this method is called on every change detection cycle and causes performance issues. This could be changed to a local variable or change this components changeDetection to onPush

vixriihi (changes requested):

vixriihi (approved):

PR #123, opened 2021-05-18
Data bank html titles 177961636

aorin:
Testable [here](https://177961636.dev.laji.fi/theme/datasets). Adds custom html titles to data bank pages.


olzraiti (approved):
:+1:

PR #124, opened 2021-05-18
Show error when requesting non existent form 178000010

aorin:
Shows an error when requesting a form that doesn't exist (for example https://178000010.dev.laji.fi/project/dsaffdsa/about).


olzraiti:
I found it a bit problematic that all pipes that use `projectForm$` have to always check that the `form` property exists - in the future if someone adds a new pipe and forgets to do check for the `form` property, an error will occur. The pipes dependent on `projectForm$` don't really make any sense, if the form is null, so it's best if `projectForm$` doesn't emit any value if the form isn't found. I wasn't entirely sure if what I just committed would be possible, so I started coding the solution to find out and it ended up working, so decided to finish it instead of requesting for changes.

Since I added quite many changes, could you check that the commit make sense?

aorin:
I think it looks good! I did some fine-tuning: I thought that it's a bit weird that notFound$ returns either boolean or string so now it returns only boolean and I removed formID from the ViewModel response since the inferface doesn't contain it.

olzraiti:
I like that! :+1:

olzraiti (approved):

PR #125, opened 2021-05-18
Fix own submissions table not showing data

aorin:
Testable [here](https://177999516.dev.laji.fi).

Steps to reproduce the bug are described in [this pivotal task](https://www.pivotaltracker.com/story/show/177999516).


olzraiti (approved):
LGTM

PR #126, opened 2021-05-20
Data bank front page changes 177961550

aorin:
Changes to dataset about page content, testable [here](https://177961550.dev.laji.fi/project/MHL.69/about).


projects/laji/src/app/+project-form/about/dataset-about/dataset-about.component.ts
```diff
@@ -1,12 +1,42 @@
...
+        collectionId: [this.form.collectionID],
+        taxonCounts: true
+      };
+      this.stats$ = this.warehouseService.warehouseQueryAggregateGet(query, [], [], 1, 1, false, false).pipe(
+        map(res => res.results[0] || {})
+      );
+    } else {
+      this.stats$ = null;
+    }
+  }
```
olzraiti, line 41: To simplify this, I think it's a fair assumption that `this.form` is defined. If it's not and an error occurs, that lets us know that there is a dataset-about component used without a form which implies a larger problem in the overall design. That way we can drop the if-else and don't have to set `stats$` to `null`, which shouldn't be done in reactive programming (RJXS). Proper way would be do make `stats$` emit an empty value.

I didn't know this beforehand, but after some reading it seems that `@Input`s can be marked required in the `@Component` `selector` like so:

```
@Component({
  selector: 'laji-dataset-about[form]',
  ...
```

We might want to add that, and start using that syntax more systemitically.

olzraiti (changes requested):

projects/laji/src/app/+project-form/about/dataset-about/dataset-about.component.ts
olzraiti: Hmm I didn't take into account that the `if (this.form?.collectionID) {` is checking also for the collectionID and not only the `form`... In practice in our problem domain here we don't necessarily have to care about that since the dataset forms always have a collectionID, but it's not a bad idea to care about it.

In that case the `stats$` shouldn't be set to `null`. Also, if we are thinking out of our problem domain and thinking in theoretical terms, the `ngOnChange` could run multiple times and in reactive programming we don't want to reinitialize the whole pipe every time, but make the same singular pipe emit the next value. So I think the cleanest solution would be wrap collectionID to an observable, and make `stats$` react to that.

aorin: The logic is now changed such that `stats$` is initialized only once. I didn't know either that it's possible to mark `@Input` required with the selector, seems useful.

olzraiti (approved):
:+1:

PR #128, opened 2021-05-25
Handle new store errors and bulk upload 178242917

vixriihi:
Show store error  messages from store if there was one and fake file loaders progress bar 
Testable at https://178242917.dev.laji.fi

### DIFF
--- a/projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
+++ b/projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
@@ -441,7 +441,7 @@ export class ImporterComponent implements OnInit, OnDestroy {
         ticker += add;
         this.current = status.processed === this.total ?
           status.processed :
-          Math.min(this.total - 1, ticker);
+          Math.min(Math.max(this.total - 1, 0), ticker);
         this.cdr.markForCheck();
       })),
       map(({errors, documents}) => rowData.map((data, idx) => {
--- a/projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
+++ b/projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
@@ -425,8 +425,10 @@ export class ImporterComponent implements OnInit, OnDestroy {
     let success = true;
     let skipped = false;
     let hadSuccess = false;
+    let ticker = 0;
     this.total = this.parsedData.length;
     this.current = 0;
+    const add = Math.min(Math.floor(this.total / 2), 50);
 
     const rowData = this.parsedData.filter(data => data.document !== null);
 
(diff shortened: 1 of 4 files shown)


projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
```diff
@@ -436,7 +438,10 @@ export class ImporterComponent implements OnInit, OnDestroy {
       publicityRestrictions
     }).pipe(
       switchMap(() => this.importService.waitToComplete('create', this.jobPayload, (status) => {
-        this.current = status.processed;
+        ticker += add;
+        this.current = status.processed === this.total ?
+          status.processed :
+          Math.min(this.total - 1, ticker);
```
olzraiti, line 444: `this.current` might end up being -1, is that intentional?

If `this.total === 0`, the latter operation of the ternary operation is executed, and `Math.min` between values `0 - 1` (this.total - 1`) and `>0` (`ticker`) will be `-1`.

projects/laji/src/app/shared-modules/spreadsheet/importer/status-cell/error-list/error-list.component.ts
```diff
@@ -65,4 +79,24 @@ export class ErrorListComponent {
     return [value];
   }
 
+  private getMessage(err: unknown): string {
+    if (typeof err !== 'object' || !Util.hasOwnProperty(err, 'message') || typeof err.message !== 'string') {
+      return this.translateService.instant('haseka.form.genericError');
+    }
+    let base = err.message;
+    if (Util.hasOwnProperty(err, 'params') && typeof err.params === 'object') {
```
olzraiti, line 87: `typeof err.params === 'object'` is `true` for an `Array` and a `null`. Are those possible types for the `err.params` in this problem domain? If not, no need for changes.

(We could add an `isObject` check to the utils that would check that it's not an array or a `null` (maybe just import that from laji-map or laji-form which have that in the utils))

olzraiti (changes requested):
Not necessary requesting changes, but some clarification requested.

projects/laji/src/app/shared-modules/spreadsheet/importer/importer.component.ts
vixriihi: `this.total` is the number of rows in the excel and there should always be some lines in it. `this.current` is only used on progress bars

projects/laji/src/app/shared-modules/spreadsheet/importer/status-cell/error-list/error-list.component.ts
vixriihi: True forgot the add the checks for the err.params

projects/laji/src/app/shared-modules/spreadsheet/importer/status-cell/error-list/error-list.component.ts
```diff
@@ -84,13 +84,8 @@ export class ErrorListComponent {
...
     }
     let base = err.message;
-    if (Util.hasOwnProperty(err, 'params') && typeof err.params === 'object') {
-      const info = [];
-      Object.keys(err.params).forEach(key => {
-        if (typeof err.params[key] === 'string') {
-          info.push(err.params[key]);
-        }
-      });
+    if (Util.hasOwnProperty(err, 'params') && typeof err.params === 'object' && err.params && Array.isArray(err.params)) {
```
olzraiti, line 87: Did you mean `Array.isArray(err.params)`? Now you are checking for an array, previously it seemed that you were checking for an object here.

olzraiti (changes requested):

olzraiti (approved):
:+1:

PR #130, opened 2021-05-25
More gis fields to frontend 178155718

MLumme:
https://178155718.dev.laji.fi/ new  geometry field to observation table, and some of then and old fields are split into multiple columns.


projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.html
```diff
@@ -2,21 +2,43 @@ <h5 *ngIf="header">
...
+        <laji-selected-field-item
+          [field]="fields"
+          [required]="required | includes:field"
+          [label]="columnsLookup[field].label | translate"
+          [idxs]="getIndexArray(fields)"
+          [len]="selected.length"
+          (toggle)="onToggle($event)"
+          (moveUp)="onMoveUp($event)"
+          (moveDown)="onMoveDown($event)"
+        ></laji-selected-field-item>
```
vixriihi, line 38: Theses two are almost the same could they be joined?

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
```diff
@@ -23,27 +23,95 @@ export class SelectedFieldGroupComponent {
...
-      if (this.columnSelector) {
-        this.columnSelector.toggleSelectedField(field);
+        if (this.columnSelector) {
+          this.columnSelector.toggleSelectedField(field);
+        }
+
+        return;
       }
+
+      this.getGisFields(field).forEach(column => {
```
vixriihi, line 36: `this.getGisFields` can return `undefined` and that would break

vixriihi (changes requested):

MLumme:
@vixriihi I simplified the components as requested.

vixriihi (approved):

PR #133, opened 2021-05-25
Fix np list displaying wrong item 178263274

aorin:
Testable here: https://178263274.dev.laji.fi

Fixes a bug described here: https://www.pivotaltracker.com/story/show/178263274


olzraiti (approved):
:+1:

PR #134, opened 2021-05-26
Change the combining default value for imports  [#177342074]

vixriihi:
Remove option to save all observations to separate documents testable at [177342074.dev.laji.fi](https://177342074.dev.laji.fi)


olzraiti (approved):

olzraiti:
This broke the tests, could you please check them?

```
1) Tools page Importing should be able send each row as document
  - NoSuchElementError: No element found using locator: By(css selector, option[value="none"])

2) Tools page Importing should be able send each row as document even with no count
  - Expected '2' to be '4', 'Selecting to skip non count rows should not effect document count'.
```

vixriihi:
yes done. This removed features that were previously tested so I now removed those tests

PR #136, opened 2021-05-26
Taxon browse images larger size button never shown as active [#177758295]

olzraiti:
https://177758295.dev.laji.fi/taxon/browse
https://www.pivotaltracker.com/story/show/177758295


esko-piirainen (approved):
ok

PR #138, opened 2021-05-27
Form robustness [#174513526]

olzraiti:
Demo https://174513526.dev.laji.fi
Task https://www.pivotaltracker.com/story/show/174513526

### Features
* Navigating away from form by any means (closing browser, in app navigation) confirms the user about leaving
* Remove document remove button from short document
* Saving document as draft bypasses warning & errors (expect schema validations)
* React & Angular errors both show a modal that comforts the user that their data is safe

### Bug fixes
* Documents with named place weren't recoverable due to incorrect temp doc key usage
* Document form was put to edit mode always

Also, some cleaning for `laji-form-document.facade` (some unused vars)


aorin (changes requested):
If I go to trip form, input geometry (and nothing else) and click "save draft", I get the "Sorry, something went wrong..." message. Is that supposed to happen?

projects/laji/src/app/shared/service/dialog.service.ts
```diff
@@ -20,24 +20,28 @@ export class DialogService {
...
+    return this.createDialog(message, true, confirmLabel, true, _default);
   }
 
-  private createDialog(message: string, showCancel?: boolean): Observable<boolean>;
-  private createDialog(message: string, showCancel: boolean, prompt: true, promptDefault?: string): Observable<string | null>;
-  private createDialog(message: string, showCancel = true, prompt = false, promptDefault?: string): Observable<boolean | string | null> {
+  private createDialog(message: string, showCancel?: boolean, confirmLabel?: string): Observable<boolean>;
+  private createDialog(message: string, showCancel: boolean, confirmLabel: string | undefined, prompt: true, promptDefault?: string): Observable<string | null>;
+  private createDialog(message: string, showCancel = true, confirmLabel?: string, prompt = false, promptDefault?: string): Observable<boolean | string | null> {
+      const initialState = {message, prompt, promptValue: promptDefault ?? ''} as any;
```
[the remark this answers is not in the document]
olzraiti: > If I go to trip form, input geometry (and nothing else) and click "save draft", I get the "Sorry, something went wrong..." message. Is that supposed to happen?

Yes, that's intentional. The form is aware only of schema + custom validations, if you look at the document save response it says: `gatheringEvent: {dateBegin: {errors: ["Date is required"]}}`. We could improve cases like this by adding stricter validation to the forms (e.g. making date begin a required field, and then trying to save as draft would display a proper validation error, because json schema validators are run against draft save). However, the currently displayed error modal is for generic errors.

> This function is starting to be quite hard to read, could it be made simpler by using an object (e.g. private createDialog(config: IDialogConfig) {})

Good idea, I'll get to it.

olzraiti:
Thanks for the suggestions, fixed the issues.

projects/laji/src/app/shared/service/dialog.service.ts
```diff
@@ -17,31 +42,33 @@ export class DialogService {
...
+    return this.createDialog<PromptConfig, string | null>({message, showCancel: true, confirmLabel, prompt: true, promptValue: _default});
   }
 
-  private createDialog(message: string, showCancel?: boolean): Observable<boolean>;
-  private createDialog(message: string, showCancel: boolean, prompt: true, promptDefault?: string): Observable<string | null>;
-  private createDialog(message: string, showCancel = true, prompt = false, promptDefault?: string): Observable<boolean | string | null> {
+  private createDialog<T extends DialogConfig, R = (boolean | (string | null))>(options: T): Observable<R> {
+    const initialState: InitialState = {showCancel: true, prompt: false,  promptValue: '', ...options};
+    if (typeof options.confirmLabel === 'string') {
+      initialState.confirmLabel = options.confirmLabel;
```
aorin, line 62: This doesn't seem to do anything since `initialState` is set to `{showCancel: true, prompt: false,  promptValue: '', ...options}`

aorin (approved):
Added a small comment but otherwise looks good :+1:

projects/laji/src/app/shared/service/dialog.service.ts
olzraiti: Good catch :+1:  I'll remove it

PR #142, opened 2021-06-01
Fix clicking "select current" on annotation form shows error

aorin:
Testable here: https://2997.dev.laji.fi/en/observation/list

Fixes the bug by removing annotationForm ViewChild that doesn't exist anymore (it can be seen in [this old version](https://bitbucket.org/luomus/laji.fi-front/src/5ce0de7b21de8c918aaa888cd69381f9925d0d2c/src/app/shared-modules/annotations/annotation-form-new/annotation-form-new.component.html) but has been since removed).


olzraiti (approved):
:+1:

PR #143, opened 2021-06-01
Revert "Revert "Revert "Auto opening selects are hiding others"""

olzraiti:
Reverception continues. This time we had to manually revert all the changes, since automatic reverting failed.


esko-piirainen (approved):
ok

PR #144, opened 2021-06-03
Sorting multi columns results in incorrect order 178281906

MLumme:
When selecting columns to show in observation table. if the order of columns is changed with arrow buttons now columns jump over all columns belonging to a multicolumn field, and don't split them as before. Branch build in https://178281906.dev.laji.fi/

### DIFF
--- a/projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
+++ b/projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
@@ -41,48 +41,32 @@ export class SelectedFieldGroupComponent {
     }
   }
 
-  onMoveUp(field: string | string[]) {
-    if (Array.isArray(field)) {
-      if (this.selected.indexOf(field[0]) > 0) {
-        field.forEach(column => {
-          this.moveUp.emit(column);
+  onMoveUp(field: string[]) {
+    const indexOfField = this.selected.indexOf(field[0]);
 
-          if (this.columnSelector) {
-            this.columnSelector.moveFieldByName(column, -1);
-          }
-        });
-      }
-      return;
-    }
+    if (indexOfField > 0) {
+      field.forEach(column => {
+        this.moveUp.emit(column);
 
-    this.moveUp.emit(field);
-
-    if (this.columnSelector) {
-      this.columnSelector.moveFieldByName(field, -1);
+        if (this.columnSelector) {
+          this.columnSelector.moveFieldByName(column, -1 * this.getFieldColumnArrayLength(this.selected[indexOfField - 1]));
+        }
+      });
     }
   }
 
-  onMoveDown(field: string | string[]) {
-    if (Array.isArray(field)) {
-      const lastSelected = this.selected.length - 1;
-      const lastField = field.length - 1;
+  onMoveDown(field: string[]) {
+    const lastSelected = this.selected.length - 1;
+    const indexOfField = this.selected.indexOf(field[field.length - 1]);
 
-      if (this.selected.indexOf(field[lastField]) < lastSelected) {
-        field.reverse().forEach(column => {
-          this.moveDown.emit(column);
+    if (indexOfField < lastSelected) {
+      field.reverse().forEach(column => {
+        this.moveDown.emit(column);
 
-          if (this.columnSelector) {
-            this.columnSelector.moveFieldByName(column, 1);
-          }
-        });
-      }
-      return;
-    }
-
-    this.moveDown.emit(field);
-
-    if (this.columnSelector) {
-      this.columnSelector.moveFieldByName(field, 1);
+        if (this.columnSelector) {
+          this.columnSelector.moveFieldByName(column, this.getFieldColumnArrayLength(this.selected[indexOfField + 1]));
+        }
+      });
     }
   }
 
@@ -105,6 +89,16 @@ export class SelectedFieldGroupComponent {
       ];
     } else {
       return [ field ];
+    }
+  }
+
+  getFieldColumnArrayLength(field: string) {
+    if (/gathering\.conversions\.(wgs84|euref|ykj)(CenterPoint)\./.test(field)) {
+      return 2;
+    } else if (/gathering\.conversions\.(wgs84|euref|ykj)\./.test(field)) {
+      return 4;
+    } else {
+      return 1;
     }
   }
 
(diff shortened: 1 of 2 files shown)


projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
```diff
@@ -41,48 +41,33 @@ export class SelectedFieldGroupComponent {
...
-          }
-        });
-      }
-      return;
-    }
-
-    this.moveUp.emit(field);
+  onMoveUp(field: string[]) {
+    const indexOfField = this.selected.indexOf(field[0]);
+    const widthOfNextGroup = this.getFieldColumnArrayLength(this.selected[indexOfField - 1]);
```
olzraiti, line 46: Minor nit: move one row below since not needed yet (less computing is good always. Also, in general in similar cases as this, computing variables only when after all sanity checks are done results in less errors, e.g. `this.getFieldColumnArrayLength` could run into an error with `undefined` (`this.selected[0 -1] === undefined`). In this particular case it doesn't run into an error, but it's generally a safer approach)

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
```diff
@@ -41,48 +41,33 @@ export class SelectedFieldGroupComponent {
...
-      const lastSelected = this.selected.length - 1;
-      const lastField = field.length - 1;
-
-      if (this.selected.indexOf(field[lastField]) < lastSelected) {
-        field.reverse().forEach(column => {
-          this.moveDown.emit(column);
+  onMoveDown(field: string[]) {
+    const lastSelected = this.selected.length - 1;
+    const indexOfField = this.selected.indexOf(field[field.length - 1]);
+    const widthOfNextGroup = this.getFieldColumnArrayLength(this.selected[indexOfField + 1]);
```
olzraiti, line 61: Minor nit: move one row below since not needed yet

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
```diff
@@ -108,6 +93,16 @@ export class SelectedFieldGroupComponent {
...
 
+  getFieldColumnArrayLength(field: string) {
+    if (/gathering\.conversions\.(wgs84|euref|ykj)(CenterPoint)\./.test(field)) {
+      return 2;
+    } else if (/gathering\.conversions\.(wgs84|euref|ykj)\./.test(field)) {
+      return 4;
+    } else {
+      return 1;
+    }
+  }
```
olzraiti, line 104: Works ok right now, but if more code is added that tries to check width for a `field` of value `undefined`, this will return `1` which could result in bugs. So a type check for the `field` is needed, and `0` should be the answer in that case.

olzraiti (changes requested):
Looking good, just some minor nits.

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
olzraiti: Actually, we don't need to hard code the array length, could you compute the values from  `this.getFieldColumnArray()`? Something like this:

```
getFieldColumnArrayLength(field: string) {
  return this.getFieldColumnArray(field)?.length || 0;
}
```

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
MLumme: Will fix this and the other one ASAP.

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
MLumme: I think your recommendation should work as-is, so I'll drop it in, and check that getFieldColumnArray has some sanity check for undefined field names.

MLumme:
@olzraiti changes pushed.

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
```diff
@@ -76,12 +79,14 @@ export class SelectedFieldGroupComponent {
...
-    if (/gathering\.conversions\.(wgs84|euref|ykj)(CenterPoint)$/.test(field)) {
+    if (!field) {
+      return;
+    } else if (/gathering\.conversions\.(wgs84|euref|ykj)(CenterPoint)/.test(field)) {
       return [
         field + '.lat',
         field + '.lon'
       ];
-    } else if (/gathering\.conversions\.(wgs84|euref|ykj)$/.test(field)) {
+    } else if (/gathering\.conversions\.(wgs84|euref|ykj)/.test(field)) {
```
olzraiti, line 89: Why the `$` check was removed from the two regexps? It's a good check, if we e.g. add a new field `gathering.conversions.wgs84.CenterPoint.SomeSubProp`, that would pass the first check. `$` means that the checked string must end there.

olzraiti (changes requested):

projects/laji/src/app/shared-modules/observation-result/selected-field-group/selected-field-group.component.ts
MLumme: Is because otherwise that function can't be used in getFieldColumnArrayLength, as the field name there has  .lat .long etc. ends. Although I could modify the regexp so that it detects either field names ending in (wgs84|euref|ykj), or it accepts lat/lon properties after it, so that the firs one could be someting like `/gathering\.conversions\.(wgs84|euref|ykj)(CenterPoin)($|.(lat|lon)$)`.

olzraiti: How about `/gathering\.conversions\.(wgs84|euref|ykj)CenterPoint(lat|lon)?$/`

MLumme: Sent changes, it now works as the old one, and it produces the column array for a member of those same columns.

MLumme: @olzraiti simplified the regexp using your suggestion.

olzraiti (approved):
:+1:

PR #146, opened 2021-06-04
Don't initialize prepopulated document with empty [#178409200]

olzraiti:
The new store has stricter validations, and thus a named place's `prepopulatedDocument` (of type `MY.document` in schema) with an empty object as the value is caught by the new stricter validators.

This PR makes it so that an empty `prepopulatedDocument` isn't added for all named places during creation/edit.

The fix is testable here (saving works, in beta it doesn't): http://178409200.dev.laji.fi/project/MHL.65/form/MHL.65/places/new
Task https://www.pivotaltracker.com/story/show/178409200


esko-piirainen (approved):
ok

PR #147, opened 2021-06-09
Named place component is more careful when calling info view methods

olzraiti:
Don't know how to replicate this, but I found from the laji.fi error logs that quite often `npClick` is undefined or `infoView` is undefined. This patch makes calling the info component methods more robust.


aorin (approved):
:+1:

PR #148, opened 2021-06-09
Remove taxon download link [#178466767]

olzraiti:
https://178466767.dev.laji.fi/en/taxon/MX.231366
https://www.pivotaltracker.com/story/show/178466767

This PR removes the download link under "Browse" section, since there isn't a download tab anymore on the observation page.


aorin (approved):
:+1:

PR #149, opened 2021-06-09
Fix SSR error in taxon identification component

aorin:
Instructions how to test this are on [the pivotal task](https://www.pivotaltracker.com/n/projects/2346653/stories/178461229/comments/224771619). 

Fixes SSR error caused by `element.getBoundingClientRect();` function being called on server.


olzraiti (approved):
:+1:

PR #157, opened 2021-06-16
Named place viewer document list on smaller-than-half-screen width

MLumme:
Got detail row to show its full contents and made the arrows indicating detail row to change correctly. Example at https://178535861.dev.laji.fi/project/MHL.1/form/MHL.1/places?activeNP=MNP.27385


olzraiti (approved):
:+1:

PR #158, opened 2021-06-17
Fix identification image modal indexing bug

Blodir:
https://www.pivotaltracker.com/story/show/178328958
https://178328958.dev.laji.fi/


esko-piirainen (approved):
ok i guess

PR #159, opened 2021-06-17
laji-form facade doesn't use timeout when reacting to data change [#178498572]

olzraiti:
Repro:

1. Goto https://beta.laji.fi/project/JX.519
2. Add a date & gathering with a geometry
3. Click date yesterday
4. Click submit very quickly right after step 3 (quicker than 3s)
5. After landing on Vihko home page, visit any other page (e.g. observations), navigate back to Vihko home page and you'll see the duplicate

This happens because:

1. At repro step 3 the `dataSub` is triggered, and it will wait 3s before continuing it's course
2. At repro step 4 the document is saved and the tmp doc is removed
3. After the 3s delay the `dataSub` saves the tmp item again.

The 3s delay has probably acted as a debouncer, so this could add a slight penalty to the form responsiveness. The form component itself has already a debouncer for string inputs, so I don't think it should be a big problem.

### DIFF
--- a/projects/laji/src/app/shared-modules/laji-form/laji-form-document.facade.ts
+++ b/projects/laji/src/app/shared-modules/laji-form/laji-form-document.facade.ts
@@ -120,7 +120,6 @@ export class LajiFormDocumentFacade implements OnDestroy {
     private personApi: PersonApi
   ) {
     this.dataSub = this.dataChange$.pipe(
-      auditTime(3000),
       mergeMap(() => this.userService.user$.pipe(take(1))),
       map(person => ({person, formData: _state.form && _state.form.formData})),
       mergeMap(data => data.formData && !_state.isTemplate ?


Blodir:
Without looking too much in depth into this... I think the logic seems a bit odd. The root issue here is that the "save" operation should override a "temp save" operation. I think using `auditTime(3000)` to implement automatic temp save makes a lot of sense.

olzraiti:
> I think using auditTime(3000) to implement automatic temp save makes a lot of sense.

Could you elaborate on this? I get that it helps a bit with performance to debounce the temp saving, but if we have to deal with bugs that it causes I'd just get rid of it. This could be solved without removing the `auditTime()`, but it'd be more complicated.

Blodir:
> > I think using auditTime(3000) to implement automatic temp save makes a lot of sense.
> 
> Could you elaborate on this? I get that it helps a bit with performance to debounce the temp saving, but if we have to deal with bugs that it causes I'd just get rid of it. This could be solved without removing the `auditTime()`, but it'd be more complicated.

Debouncing is a performance measure yes. Not a huge payoff, but that is the reason. The thing is, it's just a good practice to do it. My issue here is not that debouncing is a huge performance improvement in this specific case, but that it not being able to do it implies some fundamental issue with the underlying code. It seems like if in any scenario a temp save is triggered after a regular save with the same form state, the temp save still goes through creating a duplicate, which exposes a potential surface for bugs to appear. Furthermore if some individual interface component is accidentally not debounced it will start spamming save operations. The system just does not seem very robust, that is all.

I get that I'm being a little bit idealistic here. Removing `auditTime` fixes the problem, but I feel like the root issue is not being addressed. If refactoring is not on the table then this fix is fine.

olzraiti:
Good points! And I agree on all of that. Especially that the document facade code feels very fragile and refactoring would help. However, I think there's quite the risk that we'd break more things than fix since we don't have tests for the logic. To fix this in "ideal" way, I think we'd have to first write tests and then do the refactoring. And that's a lotta more work than just removing a single timeout operation - a lotta more work than at least I have the time for right now :)

On a side note, I'm not really a huge fan of the facade pattern - at least how we've used it. Mainly because they rely on a large singleton state, making the logic very impure. If the operations didn't rely on any state, you could use them with much more confidence.

Blodir (approved):

Blodir:
> On a side note, I'm not really a huge fan of the facade pattern - at least how we've used it. Mainly because they rely on a large singleton state, making the logic very impure. If the operations didn't rely on any state, you could use them with much more confidence.

I think the idea behind the facade pattern is good, but the implementation might need some work. The idea being that there should be a container component (in this case it's a service ie. the facade) that passes down state to dumb child components, and child components expose events to their parents. Ie. data flows down the hierarchy and events bubble up. Having "dumb" interface components makes the logic way easier to reason about,. Also it makes things like handling qparams, headers etc. much easier when all the state is in one place.

So yes, I think that having a single interface for data is good. However the architecture of the facade itself requires some thought so that it doesn't become a total mess. I do find some of the facades we have pretty hard to reason about, but I think a lot of this comes down to just general coding practices rather than the pattern itself. Like for example on `save-observations.facade` I think it's mostly pretty clear (except that I've forgotten to make the `reducer` a private method...), given this is a very simple example. You just subscribe to the output and call `load()` whenever you want new data. The `save-observations.component` can then further distribute this data down to its child components (in this case `laji-survey-box` components). In terms of the components the logic couldn't get any simpler.

Also the history of this pattern afaik is that it's a lightweight replacement to state management libraries like NgRx, Akita etc.

PR #160, opened 2021-06-18
reposition info button in select component

Blodir:
https://www.pivotaltracker.com/story/show/178569767
https://178569767.dev.laji.fi/


aorin (approved):
:+1:

PR #161, opened 2021-06-21
Lifestage filter missing values

MLumme:
Added a new method to warehouseApi to call filters-endpoint, changed metadata-select-component to use it when requested, currently used only for lifestage, but should be easily usable for any filter present in the endpoint.


olzraiti (approved):
LGTM

PR #166, opened 2021-08-17
Dependency update

olzraiti:
Updated the deps by following https://update.angular.io/ with app complexity "advanced" and migration being from 11 to 12.

1. Had to run `ng update` with `force` & `allow-dirty` flags: `ng update @angular/core@12 @angular/cli@12 @nguniversal/builders @circlon/angular-tree-component @ng-select/ng-select apollo-angular @ng-toolkit/universal --allow-dirty --force`
2. There were some breaking changes and the compiler detected some existing errors better, so I had to do the following changes:

* Use `math.div` for sass divisions
* Deleted `haseka-form-list` components, since the compiler started complaining about some stuff there and they're not used anywhere
* Added some stricter typing for components that the compiler complained about (`transfer-http-cache.interceptor.ts` and `graph-ql.service.ts`)

3. Ran `npm update`
4. Checked all items in the angular update check list (from the link in the top)


aorin:
Do you have any problem running `npm run build:prod`? I got this kind of error: "Index html generation failed. property missing ':'". (Might be related to this [issue](https://github.com/angular/angular-cli/issues/20760))

olzraiti:
Yes I get the same error. I didn't try any build scripts yet, good that you tried that out. That issue seems relevant indeed!

olzraiti:
I dug a solution from that issue discussion. Didn't test yet to actuallly build a local version to see if the solution has any side effects. We should try to remember to revert that commit in the future if the issue is fixed.

Blodir (approved):

PR #167, opened 2021-08-19
Add ely province filters 178498605

MLumme:
Adds ely-centres and provinces to area-filters. Branch build https://178498605.dev.laji.fi/observation/list


olzraiti (approved):
:+1:

PR #168, opened 2021-08-19
Project forms respect the selected lang when navigating [#178616856]

olzraiti:
https://www.pivotaltracker.com/story/show/178616856

Bug repro:
1. Go to English trip form https://laji.fi/en/vihko/home
2. Click Leave without saving

**What's the expected result?**
I'm on English page at https://laji.fi/en/vihko/home

**What's the actual result?**
I'm on Finnish page at https://laji.fi/vihko/home

Fixed all other navigation inside project form code also to translate the routes properly.


MLumme (approved):
Looks ok, although e2e-parts are outside my usual wheelhouse, so mistakes there might go unnoticed.

PR #175, opened 2021-09-09
tabs shouldn't emit when responding to content change

Blodir:
https://www.pivotaltracker.com/story/show/179494369
https://179494369.dev.laji.fi/

image


olzraiti (approved):
LGTM

PR #177, opened 2021-09-14
Import checks form id is valid 177202999

aorin:
https://www.pivotaltracker.com/story/show/177202999

Testable here: https://177202999.dev.laji.fi/project/MHL.69/import
It shows an error when uploading a file that has a wrong form id in its name (e.g. "Retkilomake (JX.519).xlsx")


olzraiti (approved):
LGTM

PR #182, opened 2021-09-20
YKJ & EMK pages moved to CMS [#179505924]

olzraiti:
https://www.pivotaltracker.com/story/show/179505924
https://179505924.dev.laji.fi/theme


MLumme:
@olzraiti Shouldn't bio-geographical province path to redirect in app-routing,modules.ts be theme/emk instead of theme/ekj? Otherwise looks fine.

olzraiti:
Yes, nice catch :)

MLumme:
Unfortunately still some issue, when trying to use https://179505924.dev.laji.fi/en/theme/emk the redirect doesn't trigger, it just shows Page Not Found, and while Swedish redirect works, the redirect removes /sv/ from the path and changes the entire site back to Finnish.

olzraiti:
Thanks, fixed that.

MLumme:
Still one more thing, which I did forget to check last time, sorry, the language reset is also present if redirect is done for /theme/ykj, present for both English and Swedish.

olzraiti:
Hmm... https://laji.fi/sv/theme/ykj and https://laji.fi/en/theme/ykj aren't usable pages really, so I don't think we need to keep backward compatibility for those. Redirecting is done just for backward compatibility. To fix that, I'd have to add the routes to the root routing like emk, but I think we should keep hacks like that as minimal as possible if there's no real benefit.

MLumme:
Aight, in that I'll mark this approved.

MLumme (approved):

PR #183, opened 2021-09-20
Rename document form cancel event with leave

olzraiti:
The `cancel` is a reserved event, and on Firefox it is called for the `document-form` component when laji-form file dialog is cancelled. On chrome the bug doesn't exist.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/cancel_event

Repro steps @ https://www.pivotaltracker.com/story/show/179412522


MLumme (approved):
Looks good and works as expected.

PR #184, opened 2021-09-20
Import observation map 177999967

aorin:
https://www.pivotaltracker.com/story/show/177999967
https://177999967.dev.laji.fi/vihko/tools/import

Adds an observation map to the 4th step of the excel importer.


olzraiti (approved):
LGTM

PR #186, opened 2021-09-20
Butterfly and bumblebee census result services improvements

MLumme:
Branch at this point contains some bugfixes for the result services which would be good to merge now, but some tasks are not done, but they can be merged in a different pull request later. Branch build here: https://179339504.dev.laji.fi/, pivotal task with bug descriptions here: https://www.pivotaltracker.com/story/show/179339504

### DIFF
--- a/projects/laji/src/app/+project-form/results/syke-insect-result/syke-insect-route-table/syke-insect-route-table.component.ts
+++ b/projects/laji/src/app/+project-form/results/syke-insect-result/syke-insect-route-table/syke-insect-route-table.component.ts
@@ -1,5 +1,5 @@
 import { Component, EventEmitter, Input, ChangeDetectionStrategy,
-Output, TemplateRef, ViewChild } from '@angular/core';
+Output, TemplateRef, ViewChild, OnInit } from '@angular/core';
 import { DatatableColumn } from '../../../../shared-modules/datatable/model/datatable-column';
 import { ExportService } from '../../../../shared/service/export.service';
 import { TranslateService } from '@ngx-translate/core';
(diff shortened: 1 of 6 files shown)


projects/laji/src/app/+project-form/results/syke-insect-result/syke-insect-route-table/syke-insect-route-table.component.ts
```diff
@@ -13,10 +13,11 @@ import { DateFormatPipe } from 'ngx-moment';
   styleUrls: ['./syke-insect-route-table.component.scss'],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class SykeInsectRouteTableComponent {
+export class SykeInsectRouteTableComponent implements OnInit, OnChanges {
   @Input() routeId: string;
   @Input() season: string;
   @Input() sorts: {prop: string, dir: 'asc'|'desc'}[] = [];
+  @Input() activeYear;
```
aorin, line 20: This is missing a type. The taxonSet variable below seems to also be missing it

aorin (changes requested):
Looks good, one small comment

projects/laji/src/app/+project-form/results/syke-insect-result/syke-insect-route-table/syke-insect-route-table.component.ts
MLumme: Fixed by removing both, they were remnants from testing and should not have been left there.

aorin: If I'm looking at correct version, taxonSet is still used in the `syke-insect-route-table.component.html` template (2nd line)

MLumme: Right you are, I missed that use, returning it and setting it's type which should be the right one.

aorin: :+1:

aorin (approved):

PR #196, opened 2021-10-19
sidebar-link require full url match for active state

Blodir:
https://www.pivotaltracker.com/story/show/179582514


Blodir:
This is intended. The idea is that currently it is unclear that you can actually click on the "Tools" link to go to tools. I suppose the ideal might be to have some sub-navigation open up to represent page hierarchy, but this is the solution for now.

olzraiti:
Ok, let's hear Mikko on this :)

PR #208, opened 2021-10-26
LajiMap pulling options from environment [#178121244]

olzraiti:
Demo: https://178121244.dev.laji.fi

The new geoserver has different address for staging & production: https://www.pivotaltracker.com/story/show/178121244

Fixed the IUCN (vir) observation map to pull the options for the map from the environment, replacing of the hard coded implementation: https://www.pivotaltracker.com/story/show/179620857

### DIFF
--- a/projects/vir/src/environments/environment.prod.ts
+++ b/projects/vir/src/environments/environment.prod.ts
@@ -80,5 +80,6 @@ export const environment = {
       en: ''
     },
   },
-  geoserver: 'https://geoserver-dev.laji.fi'
+  geoserver: 'https://geoserver-dev.laji.fi',
+  observationMapOptions: {availableOverlayNameBlacklist: []}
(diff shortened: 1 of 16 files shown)


projects/vir/src/environments/environment.prod.ts
```diff
@@ -79,5 +79,6 @@ export const environment = {
       sv: '4378',
       en: ''
     },
-  }
+  },
+  geoserver: 'https://geoserver-dev.laji.fi'
 };
```
aorin, line 84: Should this also have `observationMapOptions: {availableOverlayNameBlacklist: []}`?

olzraiti: Good catch, thanks! Fixed.

aorin (approved):

PR #211, opened 2021-11-05
Stricter typing [#180206625]

olzraiti:

There will be many similar PRs coming, read the pivotal story. This PR adds some typings and minor cleaning for the stricter tsconfig to pass.

### DIFF
--- a/projects/laji/src/app/shared/pipe/to-qname.pipe.ts
+++ b/projects/laji/src/app/shared/pipe/to-qname.pipe.ts
@@ -10,7 +10,7 @@ import { IdService } from '../service/id.service';
   name: 'toQName'
 })
 export class ToQNamePipe implements PipeTransform {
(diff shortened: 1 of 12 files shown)


Blodir:
We could maybe enable strict mode progressively since it is just a shorthand for multiple different rules: https://www.typescriptlang.org/tsconfig#strict

Blodir (approved):
There's a couple of spots where `any` was used when it could have been stricter, but it's an improvement either way 👍

olzraiti:
> We could maybe enable strict mode progressively since it is just a shorthand for multiple different rules: https://www.typescriptlang.org/tsconfig#strict

True, but then we'd have to iterate all the compiler errors multiple times to get all the strict rules to pass... My gut feeling is that iterating just once might have less overhead.

PR #213, opened 2021-11-10
Named place edit form doesn't wrap geometry to GeometryCollection [#178391265]

olzraiti:
Editing a named place now wraps the named place's geometry inside a GeometryCollection -> inconsistent data. Fixes probably multiple bugs, at least found this old pivotal ask:

https://www.pivotaltracker.com/story/show/178391265

Also after editing e.g. winter bird routes, the route list doesn't display the YKJ coordinates for the route because it's trying to show `geometry.coordinateVerbatim`, and the path is now `geometry.geometries[0].coordinateVerbatim` (after a second edit it'd be geometry.geometries[0].geometries[0].coordinateVerbatim`) 

Demo: https://178391265.dev.laji.fi/project/MHL.3


aorin (approved):
:+1:

PR #214, opened 2021-11-11
Add minimum overlap selector to protax

aorin:
Testable here: https://180230315.dev.laji.fi/theme/protax

Add minimum overlap selector to Protax form that can have values between 1-500 (it is "Minimi overlap" before translations). Also, show always link to the Protax page in the theme front page https://180230315.dev.laji.fi/en/theme


olzraiti (approved):
LGTM!

PR #220, opened 2021-11-19
Strict function types 180206625

Blodir:
Enables the `strictFunctionTypes` tsc rule.

There's a couple of `any`'s related to third party libraries that I didn't know how to type...

### DIFF
--- a/projects/laji/src/app/+taxonomy/taxon/taxon.component.ts
+++ b/projects/laji/src/app/+taxonomy/taxon/taxon.component.ts
@@ -2,7 +2,7 @@ import { ChangeDetectionStrategy, ChangeDetectorRef, Component, Inject, OnDestro
 import { ActivatedRoute, Router } from '@angular/router';
 import { combineLatest, Observable, of, Subscription, throwError } from 'rxjs';
 import { LocalizeRouterService } from '../../locale/localize-router.service';
-import { catchError, concat, delay, retryWhen, take, tap } from 'rxjs/operators';
+import { catchError, concat, delay, filter, retryWhen, take, tap } from 'rxjs/operators';
 import { Taxonomy } from '../../shared/model/Taxonomy';
 import { TaxonomyApi } from '../../shared/api/TaxonomyApi';
 import { Logger } from '../../shared/logger';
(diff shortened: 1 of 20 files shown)


aorin (changes requested):
Looks good! There is one place that I think could be improved. (Also `projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts` is hard to understand but it's not really related to this pull request)

projects/laji/src/app/+taxonomy/taxon/taxon.component.ts
```diff
@@ -110,6 +112,7 @@ export class TaxonComponent implements OnInit, OnDestroy {
 
   private initTaxon(taxonId: string): Observable<any> {
```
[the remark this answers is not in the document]
Blodir: True, fixed :)

aorin (approved):
:+1:

projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -158,7 +158,9 @@ export class SelectSubcategoriesComponent implements OnChanges {
       }
       let tmpSelectedOptions = [];
       let tmpUnselectedOptions = [];
-      const countObj = this.selectedOptions[i].filter((item: SelectOptions) => item.id !== undefined).length;
+      const countObj = this.selectedOptions[i].filter(
+        (item: SelectOptions | string) => typeof item === 'string' || item.id !== undefined
+      ).length;
```
olzraiti, line 163: BTW this broke the component, see task https://www.pivotaltracker.com/story/show/180599328

```typeof item === 'string' || item.id !== undefined```

is not the same as:
```item.id !== undefined```

I fixed this with:
```typeof item !== 'string' && item.id !== undefined```

olzraiti: Hmm, it should be the same though. Maybe some other changes here affecting this. I pushed the above as a hotfix to prod anyways - let's have a better look next week.

PR #221, opened 2021-11-19
More strict typing [#180206625]

olzraiti:
https://211119.dev.laji.fi

### DIFF
--- a/projects/laji/src/app/shared/service/id.service.ts
+++ b/projects/laji/src/app/shared/service/id.service.ts
@@ -10,7 +10,7 @@ export const DEFAULT_DOMAIN = 'http://tun.fi/';
 @Injectable()
 export class IdService {
 
-  static readonly domainMap = {
+  static readonly domainMap: Record<string, string> = {
     'luomus:': 'http://id.luomus.fi/',
     'zmuo:': 'http://id.zmuo.oulu.fi/',
     'herbo:': 'http://id.herb.oulu.fi/',
@@ -18,7 +18,7 @@ export class IdService {
     'gbif-dataset:': 'https://www.gbif.org/dataset/',
   };
 
-  static getId(value) {
+  static getId(value: any) {
     if (typeof value !== 'string' || value === '') {
       return value;
     } else if (value.indexOf(DEFAULT_DOMAIN) === 0) {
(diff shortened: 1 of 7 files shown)


projects/laji/src/app/shared/service/id.service.ts
```diff
@@ -31,14 +31,16 @@ export class IdService {
     return value;
   }
 
-  static getUri(value) {
+  static getUri(value: any) {
```
Blodir, line 34: same as above

projects/laji/src/app/shared/service/laji-api.interceptor.ts
```diff
@@ -87,14 +86,14 @@ export class LajiApiInterceptor implements HttpInterceptor {
 
     setTimeout(() => {
       if (request.body instanceof FormData) {
-        const data = [];
+        const data: any[] = [];
```
Blodir, line 89: `const data: Observable<IFileLoad>[] = [];`
(`IFileLoad` also needs to be exported)

projects/laji/src/app/shared/service/laji-api.service.ts
```diff
@@ -278,15 +278,15 @@ export class LajiApiService {
     );
   }
 
-  private removeUndefinedFromObject(obj: object) {
+  private removeUndefinedFromObject<T extends object>(obj: T): T {
```
Blodir, line 281: Doesn't this line contradict the next one? If T extends object, then there should be no need to check whether obj is an object or not?

Blodir (changes requested):

projects/laji/src/app/shared/service/id.service.ts
```diff
@@ -10,15 +10,15 @@ export const DEFAULT_DOMAIN = 'http://tun.fi/';
...
+  static readonly domainMap: Record<string, string> = {
     'luomus:': 'http://id.luomus.fi/',
     'zmuo:': 'http://id.zmuo.oulu.fi/',
     'herbo:': 'http://id.herb.oulu.fi/',
     'utu:': 'http://mus.utu.fi/',
     'gbif-dataset:': 'https://www.gbif.org/dataset/',
   };
 
-  static getId(value) {
+  static getId(value: any) {
```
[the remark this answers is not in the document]
olzraiti: Yeah I tried to refactor this first, but it wasn't very easy and I'd rather not break stuff. The generic is a good idea, I'll apply that

olzraiti:
Thanks for the review! I addressed the issues.

Blodir (approved):

PR #226, opened 2021-11-26
Parse default meta img from news content

Blodir:
If a news item has no featuredImage, parses through the html to find the first image of the article and uses that in meta tags. The meta tags should only be necessary on the server.

### DIFF
--- a/projects/laji/src/app/+news/news.component.ts
+++ b/projects/laji/src/app/+news/news.component.ts
@@ -1,4 +1,5 @@
-import { ChangeDetectionStrategy, Component, OnInit } from '@angular/core';
+import { ChangeDetectionStrategy, Component, OnInit, Inject, PLATFORM_ID } from '@angular/core';
+import { isPlatformServer } from '@angular/common';
 import { ActivatedRoute } from '@angular/router';
 import { Observable } from 'rxjs';
 import { News } from '../shared/model/News';
@@ -20,7 +21,8 @@ export class NewsComponent implements OnInit {
   constructor(private route: ActivatedRoute,
               private newsFacade: NewsFacade,
               private headerService: HeaderService,
-              private title: Title
+              private title: Title,
+              @Inject(PLATFORM_ID) private platformId
   ) {}
 
   ngOnInit() {
@@ -48,6 +50,13 @@ export class NewsComponent implements OnInit {
       this.headerService.setHeaders({
         image: news.featuredImage
       });
+    } else if (isPlatformServer(this.platformId)) {
+      const matches = news.content.match(/<img.+?src="(.*?)"/);
+      if (matches[1]?.length > 0) {
+        this.headerService.setHeaders({
+          image: matches[1]
+        });
+      }
     }
   }
 }


projects/laji/src/app/+news/news.component.ts
```diff
@@ -48,6 +50,13 @@ export class NewsComponent implements OnInit {
       this.headerService.setHeaders({
         image: news.featuredImage
       });
+    } else if (isPlatformServer(this.platformId)) {
+      const matches = news.content.match(/<img.+?src="(.*?)"/);
```
olzraiti, line 54: The regexp fails for `img` tags that have other attributes. Also it could be tighter to match exactly `img` tags with exactly `src` attribute.

For example, this would invalidly match the regexp:

``

would produce: 

`" another-tag="test`

And `` wouldn't match because of the prepending title tag.

In practice the weird tag names or attributes aren't a problem really, but it's good practice to be explicit with regexp. An attribute with quotes on the other hands is a valid problem. This regexp would be more explicit (could have some bugs still, parsing stuff with regexp can be surprisingly complicated):

`/<img +(.+ )?src="([^"]*)"/`

Here's some explanation, as regexp is infamously unreadable:
* `img +`: the tag is exactly an `img` tag, not e.g. `img-something`, followed by at least one space (`+` = at least one)
* ` (.+ )?`: before the `src` attribute there can be optionally (`?`) 1+ characters, in which case in the end there must also be a space so that the `src` attribute is exactly `src` and not e.g `something-src`
* `([^"]*)`: in this part `[^"]` means "any symbol BUT `"`. Btw, `.+?` is the same thing as `.*` :) `+` is at least one, `*` is zero or more.
```

**Review by olzraiti — CHANGES_REQUESTED** [review]
Could you provide an example page in the future? It saves time for the reviewer. I have no idea about this component and was wondering where the news content is loaded from and had to dig quite a lot to find where this component is used.

File: projects/laji/src/app/+news/news.component.ts (hunk 1 above)
Comment (Blodir) [reply projects/laji/src/app/+news/news.component.ts:54]: Both of your adversarial examples work without any issues and returns anything inside the (something-ending-with-)src="" as capture group 1. As far as weird edge cases go, the CMS API filters out any weird stuff so anything more strict than the regex I originally wrote shouldn't be necessary I think.

```
> '<img-prefixed something-src="asd" tag="test">'.match(/<img.+?src="(.*?)"/);
[
  '<img-prefixed something-src="asd"',
  'asd',
  index: 0,
  input: '<img-prefixed something-src="asd" tag="test">',
  groups: undefined
]

> '<img title="anything" src="url">'.match(/<img.+?src="(.*?)"/);
[
  '<img title="anything" src="url"',
  'url',
  index: 0,
  input: '<img title="anything" src="url">',
  groups: undefined
]
```

Blodir: Breakdown of the regex:
```<img.+?src="(.*?)"```

`<img`
`.+?` one or more of any symbol, match as few as possible
`src="` match the first occurrence of src
`(.*?)` any amount of symbols, match as few as possible
`"` match the first ending quote

olzraiti: It seems I made some mistake when testing this - I can't repro anymore this faulty match I reported:

> ``
> would produce:
> `" another-tag="test`

Sorry for the sloppiness :) Indeed the regexp works ok in this problem domain.

olzraiti (approved):

PR #231, opened 2021-12-07
Exhibition screen 1

Blodir:
* Adds a new project "exhibition-screen"
* Currently we have only 1 type of screen, which is a touchscreen slideshow
* There's a little "bug walking across the screen" animation that plays whenever the screen is idle for a while
* Note that I left some variations on the animation  in `es-animatable.ts` in case they are needed later (even though they are currently not in use)
* Looks like a small refactor intended for `taxon-identification` has also snuck in... less work to just include in this PR instead of trying to revert and redo in another branch


projects/laji/src/app/shared/service/laji-api.service.ts
```diff
@@ -201,6 +201,15 @@ export namespace LajiApi {
...
 
+export const removeUndefinedFromObject = <T extends object>(obj: T): T => {
+  return (Object.keys(obj) as (keyof T)[]).reduce((cumulative, current) => {
+    if (typeof obj[current] !== 'undefined') {
+      cumulative[current] = obj[current];
+    }
+    return cumulative;
+  }, {} as T);
+};
+
```
olzraiti, line 212: Hmm, how about moving this to `util.service.ts`? It contains more similar utils functions, and could be more easily found by other developers.

olzraiti (changes requested):
The animation stuff looks fun, and it's quite readable :) I added one minor suggestion, and once the bug-missing-its-ass-bug is solved this is good to go.

olzraiti (approved):

PR #232, opened 2021-12-13
Dependency update (ng 12 -> 13)

olzraiti:
Updated packages by the following steps (used https://update.angular.io/?l=3&v=12.0-13.0 as a checklist).

* First I updated just `nguniversal/common` `nguniversal/express-engine` from `11` -> `12`, since `npx @angular/cli@13 update @angular/core@13 @angular/cli@13` complained about those packages that multiple major version bumps can't be updated at one go.
* Updated angular packages to `13` with `npx @angular/cli@13 update @angular/core@13 @nguniversal/builders@13 @angular-devkit/build-angular@13 @swimlane/ngx-datatable --allow-dirty --force` (had to use force, forgot to save the logs sorry, it was probably something about `@ng-toolkit/universal`)
* Ran `npm update` to get minor/path level updates for all packages
* Updated `nguniversal/common` `nguniversal/express-engine` from `12` -> `13`
* Ran `npm audit fix`
* Went through all routerLinks and fixed 2 which the automatic migration failed to fix

I was slightly worried about `@ng-toolkit/universal` since it hasn't received updates and it has a peer dependency for angular 8. It's used for SSR server AFAIK, and I tested that succesfully. The repo's github issue tracker doesn't have any mentions about it breaking on newer angular versions.

I ran the e2e tests succesfully.

What do you think, instead of keeping it in a separate branch until beta freeze, how about we merge it to dev ASAP so the changes get tested as widely as possible as a side result from every other feature branch & usual development & dev.laji.fi usage? Last time we updated angular we ended up with some bugs (SSR breaking IIRC, which we have to test separately since dev pages don't use SSR).


olzraiti:
v52 has been released, so now would be optimal time to get this PR reviewed & merged to dev

Blodir (approved):
Looks good.
- I went through the angular update checklist
- Tested vir and iucn local devserver

olzraiti:
This broke the dev build:

```**ERROR** Failed to apply patch for package xlsx at path```

I removed `patches/xlsx+0.16.8.patch`, since that was the offensive patch. There wasn't any documentation what the patch was for, and I can't understand the patch file. Do you happen to have any idea what it was about @Blodir @aorin ? Viltsu had added the patch.

Apparently the patch files are made by directly manipulating a file under `node_modules`, and then running `patch-package` that creates the patch file from the changes made. In the future, if such patches are needed, we should properly document what the patch does in the commit that adds it (related github issues, even a file diff). We might need to test that excel importing/creation works correct...

btw ci broken even after removing the fix, didn't find yet why from the log. I'll have a look tomorrow.

olzraiti:
Ok, the patch files are just diff files so I created a new identical one for `xlsx@0.17.4` which is the new version we are using. Still no idea what the patch does or why... If something comes up with excel related code, we can downgrade `xslx` back to the old version.

CI is broken because `tslint` is deprecated and the lint xml result generation fails. So, tomorrow I'll spend the refactoring friday by migrating to `eslint`.

PR #233, opened 2021-12-13
Project form latest docs fix [#180618581]

olzraiti:
Task: https://www.pivotaltracker.com/story/show/180618581
Demo: https://180618581.dev.laji.fi/project/MHL.51

The project form page is supposed the latest docs in the sidebar (or if the form is a mobile form, on the about page). It worked ok for local tmp docs, but the remote docs showed only the form's docs if they happened to be very recent (in the 10 docs you have last reported). The fix was to pass the formID to the query being made.


MLumme (approved):
Seems ok as far as I can tell.

PR #236, opened 2021-12-16
Various vir (PAP) changes

Blodir:
In this PR (iirc, might be more):
* moving to the new /warehouse/downloads endpoint
* fix my downloads showing everyone's downloads
* fix datatable exports
* update some arcgis cms ids


projects/vir/src/app/+usage/component/data-table/data-table.component.ts
```diff
@@ -38,7 +39,7 @@ type TableType = 'downloads'|'people'|'user'|'userKeys'|'apiKeys';
           </laji-datatable>
       </div>
       <ng-template let-value="value" let-row="row" let-sort="sortFn" #downloadFileTpl>
-        <a [href]="'/api/file-download?id=' + value">{{ ('download.' + row.downloadType) | translate }}</a>
+        <a [href]="'/api/warehouse/download/secured/' + value + '?personToken=' + getPersonToken()">{{ ('download.' + row.downloadType) | translate }}</a>
```
olzraiti, line 42: Using function calls in template code should be avoided, how about just bind the person token to an instance variable?

projects/vir/src/app/component/footer/footer.component.ts
```diff
@@ -17,6 +17,7 @@ const informationWhitelist = [
   // Services and instructions
   '803', // tietovarasto
   '806c1', '806', // API / technical
+  '5138',
```
olzraiti, line 20: Everything else being commented here, maybe add a comment for this also?

olzraiti (changes requested):
Few nits, otherwise looking good

olzraiti (approved):
Thanks!

PR #237, opened 2021-12-22
Fix document form facade named place observable [#180717623]

olzraiti:
Task https://www.pivotaltracker.com/story/show/180717623
Bug demo: https://beta.laji.fi/project/JX.519/form/JX.283171
Fix demo https://180717623.dev.laji.fi/project/JX.519/form/JX.283171

The named place pipe crashes silently on a null pointer error, resulting in wrong error message being displayed.


aorin (approved):
:+1:

PR #238, opened 2021-12-29
Fix prefilled forms can be only used once

aorin:
Pivotal task: https://www.pivotaltracker.com/story/show/180748974
Testable here: https://180748974.dev.laji.fi/vihko/templates

Fixes a bug where prefilled forms can only be used once without logging out or refreshing the page. The problem was that the `documentService.removeMeta` function modifies the document object directly so we need to make a copy of it before calling it.


olzraiti:
To prevent this in the future, how about making `documentService.removeMeta` pure so that it does the cloning? I checked that all usages of the method rebind the result to a new variable, expect one (in `projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts`), but that one also does clone the document before the method usage so should be easy to refactor.

olzraiti (approved):
:+1:

PR #240, opened 2022-01-12
Latest docs bug on project form [#180846101]

olzraiti:
Task: https://www.pivotaltracker.com/story/show/180846101
Demo: https://180846101.dev.laji.fi/project/MHL.3/form

The bug being fixed here is after submitting a new document for a project form, the latest docs in the sidebar will show all documents instead of just that form's documents. I added a `setFormID()` method to the latest docs facade, so that the `update()` method can be called without resetting the formID (previously the `update()` method updated the formID for the facade).


MLumme (approved):
Looks fine.

PR #242, opened 2022-01-12
Fix observation table sorting

Blodir:
https://www.pivotaltracker.com/story/show/180765737
https://180765737.dev.laji.fi/


aorin (approved):
Looks good

PR #243, opened 2022-01-13
Taxon identification fix 180717763

Blodir:
- Refactored sub-mainrank finding logic (the closest main rank that is below the rank of the taxon)
- Fixed behavior when taxon rank is below the lowest mainrank
- Fixed old data source being potentially disconnected more than once
- Made sure the loader disappears after the sub-mainrank has been received regardless of if there are any child species

https://www.pivotaltracker.com/story/show/180717763
https://180717763.dev.laji.fi/taxon/MX.5073946/identification


olzraiti (approved):
:+1:

PR #244, opened 2022-01-14
Move informal group svg to cdn

olzraiti:
Moved informal taxon group icons to cdn for easier development (for example no need to define dev/prod static img site for them on laji-form).

Demo https://220114.dev.laji.fi/ for example search for species and the correct svg images will be shown for the results.


Blodir (approved):
Looks good

PR #250, opened 2022-02-02
Graphql fix

Blodir:
updated schema: https://bitbucket.org/luomus/laji-api-graphql/commits/499e43235064129c833f7ae60a26eea88257c41f
updated schema, fixed graphql query duplicate private key bug, fixed collectionAndRecordQuality bug, added e2e test: https://github.com/luomus/laji/compare/graphql-fix-123

note about collectionAndRecordQuality: it's now passed as a string to avoid issues with delimiters (this field uses ; instead of ,)

https://123.dev.laji.fi/


esko-piirainen (approved):
Well no code style comments and can't spot bugs ..   Can't say much else

PR #254, opened 2022-02-04
Add hard coded form category order list [#181151175]

olzraiti:
https://www.pivotaltracker.com/story/show/181151175
https://181151175.dev.laji.fi/save-observations

The goal being that bird atlas form is shown first. I added a hard coded order list, which is read so that forms not found in the list are ordered by ID after the ones from the order list.


MLumme (approved):
Looks fine.

PR #256, opened 2022-02-08
PAP ELY search

Blodir:
https://www.pivotaltracker.com/story/show/180597998
https://www.pivotaltracker.com/story/show/180598385

https://viranomaiset-dev.laji.fi/observation/list
(to be approved)

### DIFF
--- a/projects/laji/src/app/+observation/form/observation-form.component.ts
+++ b/projects/laji/src/app/+observation/form/observation-form.component.ts
@@ -126,19 +126,15 @@ export class ObservationFormComponent implements OnInit, OnDestroy {
   private _query: WarehouseQueryInterface;
 
   virFilterShortcutQueryParams = {
-    administrativeStatusId: `
-      MX.finlex160_1997_appendix4_2021,MX.finlex160_1997_appendix4_specialInterest_2021,
-      MX.finlex160_1997_appendix2a,MX.finlex160_1997_appendix2b,MX.finlex160_1997_appendix3a,
-      MX.finlex160_1997_appendix3b,MX.finlex160_1997_appendix3c,MX.finlex160_1997_largeBirdsOfPrey,
-      MX.habitatsDirectiveAnnexII,MX.habitatsDirectiveAnnexIV,MX.birdsDirectiveStatusAppendix1,
-      MX.birdsDirectiveStatusMigratoryBirds`,
-      redListStatusId: 'MX.iucnCR,MX.iucnEN,MX.iucnVU,MX.iucnNT',
-      countryId: 'ML.206',
-      time: '1990-01-01%2F',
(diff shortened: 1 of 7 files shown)


projects/laji/src/app/+observation/form/observation-form.component.ts
```diff
@@ -125,6 +125,22 @@ export class ObservationFormComponent implements OnInit, OnDestroy {
...
+      MX.finlex160_1997_appendix3b,MX.finlex160_1997_appendix3c,MX.finlex160_1997_largeBirdsOfPrey,
+      MX.habitatsDirectiveAnnexII,MX.habitatsDirectiveAnnexIV,MX.birdsDirectiveStatusAppendix1,
+      MX.birdsDirectiveStatusMigratoryBirds`,
+      redListStatusId: 'MX.iucnCR,MX.iucnEN,MX.iucnVU,MX.iucnNT',
+      countryId: 'ML.206',
+      time: '1990-01-01%2F',
+      collectionAndRecordQuality: `PROFESSIONAL:EXPERT_VERIFIED,COMMUNITY_VERIFIED,NEUTRAL,UNCERTAIN;
+      HOBBYIST:EXPERT_VERIFIED,COMMUNITY_VERIFIED,NEUTRAL;AMATEUR:EXPERT_VERIFIED,COMMUNITY_VERIFIED;`,
+      individualCountMin: 0,
+      coordinateAccuracyMax: 1000
```
olzraiti, line 141: These should be indented one level less.

projects/laji/src/app/+observation/observation.component.ts
```diff
@@ -40,7 +41,7 @@ export class ObservationComponent extends AbstractObservation implements OnInit,
 
   ngOnInit() {
     this.observationFacade.emptyQuery = {
-      _coordinatesIntersection: 100
+      _coordinatesIntersection: environment.type === 'vir' ? 0 : 100
```
olzraiti, line 44: Hmm... Could we put this logic into the `vir` env file instead of doing a hard check for the env name here? The hard check isn't really that bad, but the env variable -way would be more consistent with our existing code. Relying on an env file is more agnostic approach also.

So `projects/vir/src/environments/environment.ts` would have `observationCoordinatesIntersection` at here we would check pull the value from the env. Similar to how `observationMapOptions` is implemented in the env files.

olzraiti (changes requested):

projects/laji/src/app/+observation/observation.component.ts
olzraiti: Maybe a more extendable solution would be to have `observationQuery` var instead of `observationCoordinatesIntersection`, and `this.observationFacade.emptyQuery` would spread it into the var (`{...bla, ...(env.observationQuery || {})}`)

Blodir: hmm well i already did the simple version of this, it's easy to extend if it is ever needed

olzraiti (approved):
Thanks, looking good! Is this waiting still for approval or is it ok to merge to beta?

PR #261, opened 2022-02-17
Import laji-map typings directly from laji-map

olzraiti:
Previously we had to to handle laji-map typings manually, so that laji-map wouldn't be imported for SSR (where it doesn't work). Nowadays typescript can handle importing only typings from a library, so importing the typings in SSR is safe. So, in this PR I made the imports use laji-map directly, instead of having to define the typings by hand everytime we need to import some types from laji-map.

I tested that building SSR works.


olzraiti:
Sadly, this had to be reverted as SSR broke :( I tested that the server started but apparently hadn't tested actually visiting the front page. It seems that the import system isn't robust enough to be trusted yet.

PR #263, opened 2022-03-01
Lightweight GIS download

aorin:
Adds GIS formats to lightweight download. 

Testable here https://179718878.dev.laji.fi/observation/list and here https://179718878.viranomaiset-dev.laji.fi/observation/list
(Filter results -> Click Download -> Select .shp or .gpkg as format)


projects/laji/src/app/shared/service/export.service.ts
```diff
@@ -129,4 +120,31 @@ export class ExportService {
...
+      type = this.zipMimeType;
+    } else {
+      type = this.txtMimeType;
+    }
+
+    const data: Blob = new Blob([buffer], {
+      type
+    });
+
+    return data;
```
olzraiti, line 144: The blob could be just returned right away, the `data` variable has no purpose

olzraiti (changes requested):
Some very minor nits, otherwise looks OK!

PR #267, opened 2022-03-02
Map print & squaregrid parameters  [#181266022]

olzraiti:
Adds the URI param `print` and `squaregrid` to the `/map` service. PDF print button was refactored to it's own component.

https://181266022.dev.laji.fi/map?print=true&gridsquare=666:333&layers=maastokartta,atlasGrid
https://www.pivotaltracker.com/story/show/181266022


MLumme (approved):
Seems fine, accepted.

PR #268, opened 2022-03-04
Enable 'noImplicitAny' and 'strictInputAccessModifiers` [#180206625]

olzraiti:
https://220304.dev.laji.fi

`strictInputAccessModifiers` came free, didn't raise any compiler issues.

### DIFF
--- a/projects/laji/src/app/+project-form/form/named-place-linker/named-place-linker.component.ts
+++ b/projects/laji/src/app/+project-form/form/named-place-linker/named-place-linker.component.ts
@@ -116,6 +116,7 @@ export class NamedPlaceLinkerComponent implements OnInit, OnDestroy {
       this.loading = false;
       this.translate.get('np.linker.success').pipe(take(1)).subscribe(msg => this.toastsService.showSuccess(msg));
       this.reloadSubmissions$.next();
+      return EMPTY;
     });
   }
 }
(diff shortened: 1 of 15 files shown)


projects/laji/src/app/+project-form/form/named-place-linker/named-place-linker.component.ts
```diff
@@ -116,6 +116,7 @@ export class NamedPlaceLinkerComponent implements OnInit, OnDestroy {
       this.loading = false;
       this.translate.get('np.linker.success').pipe(take(1)).subscribe(msg => this.toastsService.showSuccess(msg));
       this.reloadSubmissions$.next();
+      return EMPTY;
```
aorin, line 119: Since this is inside rxjs `subscribe`, I don't think this function should return anything

projects/laji/src/app/shared-modules/document-viewer/document-annotation/document-annotation.component.ts
```diff
@@ -243,6 +243,7 @@ export class DocumentAnnotationComponent implements AfterViewInit, OnChanges, On
             return hasHighlight;
           }
           hasHighlight = this.shouldOnlyShowHighlighted(subLevel, highlight);
+          return undefined;
```
aorin, line 246: `forEach` function shouldn't return anything so maybe this could be refactored like this
```
if (!hasHighlight) {
    hasHighlight = this.shouldOnlyShowHighlighted(subLevel, highlight);
}
```

olzraiti:
Thanks! Fixed those.

aorin (approved):
:+1:

PR #269, opened 2022-03-07
Label designer improvements

aorin:
Various improvements to label designer
https://www.pivotaltracker.com/story/show/173725824

- Decreased minimum label size (https://www.pivotaltracker.com/story/show/173725824)
- Added domain to list of fields in the excel column mapping (https://www.pivotaltracker.com/story/show/178811314)
- Available fields are reset when clear data is clicked (https://www.pivotaltracker.com/story/show/178811281)
- Fixed a bug where fields selector shows "Text" instead "add..." (https://www.pivotaltracker.com/story/show/180795932)


projects/label-designer/src/lib/label-designer.helper.ts
```diff
@@ -42,4 +42,19 @@ export class LabelDesignerHelper {
...
+    }
+    const keys = new Set<string>();
+    fields2.forEach(field => keys.add(field.field));
+    let result = true;
+    fields1.forEach(field => {
+      if (result && !keys.has(field.field)) {
+        result = false;
+      }
+    });
+    return result;
```
olzraiti, line 58: Could be more simple with declarative approach, e.g.:

```
return fields1.every(f=> keys[f.field]))
```

That would also prevent excessive iteration, as the `every` would stop when a non matching item is found.

Continuing with declarative style, the `keys` could also be declared from `fields2` using reduce. IMO declarative style is more readable than line-by-line code execution. But that's just my opinion, take it as an idea instead of a change requirement :)

**EDIT:** I just noticed that this code block was copy-pasted from elsewhere, so no need to change anything as this isn't new code. I'll leave this comment here anyways. If you have time, you can improve the code of course!

projects/label-designer/src/lib/label-designer/label-settings/available-fields/available-fields.component.ts
```diff
@@ -10,23 +14,35 @@ import { FieldType, ILabelField } from '../../../label-designer.interface';
...
+  selectedIdx: number;
+
+  ngOnChanges() {
+    this.availableFieldsWithIdx = (this.availableFields || []).map((field, idx) => ({
+      ...field,
+      idx: idx
+    }));
+    this.selectedIdx = this.availableFieldsWithIdx.filter(
+      field => field.field === this.value && !(field.type == this.fieldType.qrCode || field.type == this.fieldType.text)
+    )[0]?.idx;
```
olzraiti, line 37: Use `findIndex()`:

```
this.availableFieldsWithIdx = this.availableFields?.findIndex(field =>
  field.field === this.value && !(field.type == this.fieldType.qrCode
  || field.type == this.fieldType.text)
);
```

(didn't test that but should do the trick)

olzraiti (changes requested):
Looks good! I left one change requirement.

projects/label-designer/src/lib/label-designer/label-settings/available-fields/available-fields.component.ts
aorin: Changed it to use `findIndex`. Still couldn't get rid of the variable `this.availableFieldsWithIdx` because it's needed in the template but it looks a little better now

olzraiti (approved):
Thanks, looks good!

PR #270, opened 2022-03-07
Add onlyNonStateLands filter to observation search

aorin:
Testable here: https://180707580.dev.laji.fi/observation/list?onlyNonStateLands=true


olzraiti (approved):
LGTM!

PR #271, opened 2022-03-17
Bird atlas release

Blodir:


olzraiti (changes requested):
I reviewed `projects/laji` part. I left a comment for the coordinate service. The biggest change seemed to be that `PlatformService` moved from `shared` to `shared-modules`. What's the reasoning behind that? 

The way I see the difference between the folders:
* `shared` is for code that is provided in the `shared.module` OR provided by root.
* `shared-modules` contains "public" modules that can be imported to any other modules

`PlatformService` is provided by root, so following this categorization it should stay in `shared`?

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
```diff
@@ -1,9 +1,156 @@
...
+  private loadMap$ = new BehaviorSubject<string>(undefined);
+
+  data$ = this.route.paramMap.pipe(
+    tap(() => {
+      this.breadcrumbs.setBreadcrumbName(BreadcrumbId.GridInfo, undefined);
+      this.loading = true;
+      this.cdr.markForCheck();
+    }),
+    switchMap(params => this.atlasApi.getGridElement(params.get('id'))),
+    catchError(() => of(undefined)),
```
aorin, line 38: I think it's little confusing to have `catchError` in the middle, how about this kind of structure
```
...,
switchMap(params => this.atlasApi.getGridElement(params.get('id')).pipe(
    map(<map function>),
    tap(<tap function)
),
catchError(() => ({
      elem: undefined,
      rows: [],
      status: 404
    })
),
tap(() => {
     // this is needed in both cases
    this.loading = false;
    this.cdr.markForCheck();
}));
```

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
```diff
@@ -1,9 +1,156 @@
...
+
+  ngAfterViewInit(): void {
+    this.loadMap$.pipe(
+      takeUntil(this.unsubscribe$),
+      tap(() => { if (this.map) { this.map.destroy(); } }),
+      filter(d => d !== undefined)
+    ).subscribe(ykj => {
+      setTimeout(() => { // yield control to renderer before accessing lajiMapElem
+        this.zone.runOutsideAngular(() => {
+          this.map = new LajiMap({
```
aorin, line 120: Why the map is created every time when the data changes and why the laji map component is not used?
`<laji-map [data]="data" [options]="mapOptions" [visible]="visible"></laji-map>`

projects/bird-atlas/src/app/core/scroll-position.service.ts
```diff
@@ -0,0 +1,34 @@
...
+
+@Injectable()
+export class ScrollPositionService {
+  private urlBeforeNavigation: string;
+  private urlToScrollPosition: {[url: string]: number} = {};
+
+  constructor(router: Router, @Inject(PLATFORM_ID) platformId) {
+    if (!isPlatformBrowser(platformId)) { return; }
+    this.urlBeforeNavigation = router.url;
+    router.events.pipe(
```
aorin, line 13: this empty pipe could be removed

projects/bird-atlas/src/app/+species/species-info/species-info.component.ts
```diff
@@ -13,19 +19,64 @@ import { ApiService } from '../../core/api.service';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class SpeciesInfoComponent {
-  taxon$: Observable<Taxon> = this.route.paramMap.pipe(
-    switchMap(params => this.api.getTaxon(params.get('id'), {}, 'multi'))
+  data$ = this.route.paramMap.pipe(
```
aorin, line 22: this could also have explicit type declaration

projects/bird-atlas/src/app/app.component.ts
```diff
@@ -25,7 +34,22 @@ import { TranslateService } from '@ngx-translate/core';
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class AppComponent {
-  constructor(translate: TranslateService) {
+  breadcrumbs$: Observable<IBreadcrumb[]> = this.breadcrumbs.breadcrumbs$.pipe(tap(() => this.cdr.detectChanges()));
```
aorin, line 37: I'm just wondering if `this.cdr.detectChanges()` is needed here. Shouldn't the changes be always detected when using async pipes?

aorin (changes requested):
I looked through the `projects/bird-atlas` part and generally looks very good. There were some things that caught my attention (see the comments below).

projects/laji/src/app/shared/service/coordinate.service.ts
```diff
@@ -1,6 +1,38 @@
...
+        [latStart, lonStart],
+        [latStart, lonEnd],
+        [latEnd, lonEnd],
+        [latEnd, lonStart],
+        [latStart, lonStart],
+      ] as [string, string][]).map((latLng: [string, string]) => convertYkjToWgs(latLng).reverse())],
+      coordinateVerbatim: origLat + ':' + origLng
+    }
+  };
+};
```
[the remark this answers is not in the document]
Blodir: The way I see it, it's confusing that there's stuff in the `shared` directory that is not declared/provided in `SharedModule`. It came up when I was trying to evaluate whether a directive that imported platformservice would require `SharedModule`. There's really no choice but to check the code of platformservice and everything that's injected into it... Imo it would be nice if the directory hierarchy and module system in combination gave a good idea of the dependencies of each component/service/w.e. In fact I think `providedIn: 'root'` might be a little bit of an antipattern too because of this...

I'm not married to the idea of moving platform service btw, but I think it's nice being able to tell that it's not a part of shared module from looking at the filepath alone.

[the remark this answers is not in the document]
Blodir: > Why were these moved away from the class? Having such static methods is fine and could be a reasonable option for all the functions that the coordinate service provides as it doesn't rely on any other services. but I think we should have either everything in the service class, OR a utility file that exports everything flat, but having both makes it unpredictable how the functions should he imported.

I prefer using js modules to namespace stuff over classes with static methods (the OOP way).
1. if the class has non-static methods then static methods are unnecessarily cluttering the class definition
2. class notation implies that there's shared state between the methods and static methods are basically just a workaround for pure OOP (or close to) languages (ie. not js)
2.2 to clarify my hot take further: the entire point of classes is that you can instance them. If I see a class, it's my immediate assumption that you are supposed to instance it. Static methods break this pattern.
3. static methods introduce an unnecessary level of indentation
4. functions can still be namespaced if necessary without relying on class notation (which comes with the aforementioned baggage of being associated with actual class instancing) or js modules: just define it inside a regular js object eg. **https://github.com/luomus/laji/blob/bird-atlas/projects/exhibition-screen/src/animations/math.ts**

Ie. my personal preference is to never use static methods in ts/js at all and rather leave the class definitions for things that actually use the functionality that the class keyword provides: instancing.

`CoordinateService` is particularly suspicious since it doesn't have any shared state despite all methods being non-static. There's no reason this needs to be a service. I can convert the entire class into utility functions if that's okay with you?

projects/bird-atlas/src/app/app.component.ts
Blodir: Yeah I don't know why this one is here, seems to work without it. There's a similar case here: https://github.com/luomus/laji/blob/bd8011c53fa90833f32fe7a20bde6da144c0b073/projects/bird-atlas/src/app/+species/species-info/species-info.component.ts#L53, but in this case the detectChanges was actually necessary and I have no idea why. This detectChanges fixes a bug where the spinner would not disappear after navigating to a species page -> another page -> back to the same species page.

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
```diff
@@ -1,9 +1,156 @@
...
+  changeDetection: ChangeDetectionStrategy.OnPush
 })
-export class GridInfoComponent {
+export class GridInfoComponent implements AfterViewInit, OnDestroy {
+  @ViewChild('lajiMap', { static: false }) lajiMapElem: ElementRef;
+
+  private unsubscribe$ = new Subject<void>();
+  private loadMap$ = new BehaviorSubject<string>(undefined);
+
+  data$ = this.route.paramMap.pipe(
```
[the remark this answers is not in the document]
Blodir: I'll add the type declaration 👍 

I get what you mean about the chunky class variable definition being a bit weird, but moving it to the constructor doesn't really change much other than adding another level of identation imo :/

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
Blodir: 👍

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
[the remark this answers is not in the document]
Blodir: Actually I take it back it's nicer to move stuff to the constructor so that you can see all the class variables with one glance :)

projects/laji/src/app/shared/service/coordinate.service.ts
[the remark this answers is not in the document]
olzraiti: Thanks for elaborating! Your points are valid and in many (and probably most) cases are the best way. I think however that there are valid cases for static methods: if you have a service that can't be static, but it has some static utility methods, putting it in the same namespace can make sense if it's needed elsewhere (makes me wonder if a private static method would ever be useful...). But even that is a matter of taste, and I'm not sure it applies for all such cases.

> CoordinateService is particularly suspicious since it doesn't have any shared state despite all methods being non-static. There's no reason this needs to be a service (or a class). I can convert the entire class into utility functions if that's okay with you?

Sounds good! It's just a collection of static utility functions so this would make a lot of sense.

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
Blodir: `LajiMapComponent` requires a lot of stuff that is not included in `LajiMapModule` and I didn't spend time refactoring it all...
eg.
* environment props has to include apiBase, loginUrl, systemID (note that bird-atlas has no login feature)
* bunch of other userservice dependencies
* provider for logger service

projects/laji/src/app/shared/service/coordinate.service.ts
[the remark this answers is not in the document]
olzraiti: > The way I see it, it's confusing that there's stuff in the shared directory that is not declared/provided in SharedModule. It came up when I was trying to evaluate whether a directive that imported platformservice would require SharedModule. There's really no choice but to check the code of platformservice and everything that's injected into it... Imo it would be nice if the directory hierarchy and module system in combination gave a good idea of the dependencies of each component/service/w.e. In fact I think providedIn: 'root' might be a little bit of an antipattern too because of this...
> 
> I'm not married to the idea of moving platform service btw, but I think it's nice being able to tell that it's not a part of shared module from looking at the filepath alone.

Yeah, the root issue seems to be that our folder structure doesn't reflect the app structure very well... The stuff provided in `root` don't fit in either folder really. If you put them in the `shared` you might think that it's provided in the shared module. If you put it in the `shared-modules`, you might think that it's included in some module and that's also confusing. Neither place really makes sense. We should refactor the module structure... So, I don't have a strong opinion where the platform service should reside since there isn't any correct answer. Making changes that don't solve anything doesn't really help and makes following git changes more difficult, and there's already a merge conflict and the beta merge might revolt in a lot of those. In fear of merge conflicts, I'd vote on reverting the moving of the platform service.

Maybe we should have a dev-meeting about what are the current key issues that could be refactored in laji repo. Usually summer has been good time for refactoring stuff :)

We should discuss this deeper at some other point, but I'm just writing down what I'm thinking the structure could be look like:
* `modules`: current `shared-modules`
* `modules/shared` current `shared`
* `root` for stuff provided by root? You might be right that it's an antipattern to provide in the root, but bloating the shared module to include everything is maybe also not the optimal solution.

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
aorin: Ok, that makes sense

projects/bird-atlas/src/app/app.component.ts
aorin: That case in `species-info.component`is weird but we have had similar cases before... The problem is that when navigating back to species page, the observable emits a value too soon (as all the values are in cache). I'm not sure exactly but I guess that causes `isLoading` changing in the middle of Angular change detection cycle and that's something that shouldn't happen. We have usually fixed this kind of problems by adding `delay(0)` before the final `tap` instead of `this.cdr.detectChanges`

projects/laji/src/app/shared/service/coordinate.service.ts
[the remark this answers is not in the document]
Blodir: `PlatformService` moved to `/root`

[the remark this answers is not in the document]
Blodir: Oops, forgot about that! Ok, `shared/service/coordinate.service` is now `root/coordinate-utils.ts` and exports bare functions. Perhaps a further consideration would if the `/root` would be renamed to `/no-module` (or perhaps `/moduleless`, or did I just play too much elden ring?) or some such, but I think '/root' is a decent enough place for the time being.

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
```diff
@@ -1,9 +1,172 @@
...
+          title: `${d.elem.name} ${d.elem.coordinates} | ${this.translate.instant('ba.header.title')}`
+        });
+        this.loadMap$.next(d.elem.coordinates);
+      }),
+      catchError(() => of({
+        elem: undefined,
+        rows: [],
+        status: 404
+      })),
+      tap(() => this.loading = false)
```
aorin, line 114: You might need to change this to `delay(0), tap(() => this.loading = false)` (or add `cdr.detectChanges()`) since it looks like the same can happen here than in the `species-info.component` where the spinner spins indefinitely

aorin (changes requested):
Looks good! When I tried to test the grid page, it seems to broken:
https://tuloksia-dev.lintuatlas.fi/grid/678:333
Do you know what's the problem there?

projects/bird-atlas/src/app/+grid/grid-info/grid-info.component.ts
Blodir: Looks like AtlasApi no longer returns the coordinates field. @MLumme can this field make a comeback, or perhaps the id field should be returned without the tun.fi part?

Blodir: Added the `detectChanges()`. Imo using `detectChanges()` rather than `delay(0)` feels more natural and the role of it is a perhaps a bit less mysterious.

Blodir: > Looks like AtlasApi no longer returns the coordinates field. @MLumme can this field make a comeback, or perhaps the id field should be returned without the tun.fi part?

Should be safe to merge since this is apparently a bug in the backend.

aorin: Ok thanks! I think this is ready then

aorin (approved):

projects/laji/src/app/shared/service/coordinate.service.ts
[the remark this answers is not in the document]
olzraiti: Thanks, lookin' good! Yeah it could help if the naming somehow would implicate that it's not included in any module, but indeed it's already better so let's merge this :)

olzraiti (approved):

PR #272, opened 2022-03-30
Limit gis downloads

aorin:
https://www.pivotaltracker.com/story/show/181521035

Shows a warning when making a data request if there are too many observations for GIS download (testable here https://181521035.dev.laji.fi/observation/list).
Disables GIS option when downloading a file if there are too many observations (testable here https://181521035.viranomaiset-dev.laji.fi/usage/my-downloads).


olzraiti (approved):
:+1:

PR #276, opened 2022-04-06
Species count query changes

aorin:
Change species count query so that the count is the same than the total count in the species table

https://181420427.dev.laji.fi/observation/species


esko-piirainen (approved):
ok

PR #283, opened 2022-04-26
Polygon search [#181264927]

olzraiti:
Demo: https://181264927.dev.laji.fi/observation

Docs: 
https://www.pivotaltracker.com/story/show/181264927
https://wiki.helsinki.fi/display/luomusict/Polygon+search (section "How to implement to laji.fi" is most relevant)

Adds the possibility to draw (or enter in string format) a polygon to the observation map.

Technical summary:

1. The polygon is registered to the warehouse, and the returned polygon id is used as the warehouse occurrence search filter.
2. if the coordinates intersection is undefined, 0 will be used as the default ~(in contrast to the bounding box filter, which uses 1 as the default)~ EDIT: 0 is the default for bounding box now also
3. User must be logged in to register a polygon

The task is approved and thus ready for review, but not yet ready for production.


Blodir (changes requested):
Looks good, however I can't say I fully understand what is happening at every part of this PR. I also have a few thoughts I posted as comments.

Blodir (approved):

PR #286, opened 2022-05-05
Bird atlas grid index map

Blodir:
Users can now navigate to a grid square using a map view.
* Grid-index is expanded to include two sub-components (map and table), while the parent component handles data flow and tabs
* Some laji-ui base styles bled over to laji-map controls. I'm not entirely sure if there's a good way to prevent this, but in this PR those styles are simply overwritten in `laji-map-overwrites.scss`. Currently thinking about perhaps reducing the level of necessary css qualification to the base styles by 1 level by removing the `.main-view` class and just overwriting some styles on the navbar to compensate.
* Gradient color choice may still change (in which case I will commit directly to dev/bird-atlas branch)
* Note that the `color-math.ts` utility can be used to generate color gradients based on linear hsl interpolation, but it might be wise to remove calls to it once the gradient choice is finalized (and just inline the result colors) to slightly improve performance.


projects/bird-atlas/src/app/+grid/grid-index/grid-index-table/grid-index-table.component.ts
```diff
@@ -0,0 +1,85 @@
...
+    name: this.translate.instant('ba.grid-index.birdAssociationArea'),
+    resizeable: false,
+    sortable: true,
+    width: 350
+  }];
+  filteredRows$ = new Subject<DatatableRow[]>();
+  datatableClasses = datatableClasses;
+
+  private search$ = new BehaviorSubject<string>(undefined);
+  private rows: DatatableRow[] = [];
```
aorin, line 50: I'm wondering why any type is used here (`type DatatableRow = any` ) when more specific type could be used (AtlasGrid)

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
```diff
@@ -0,0 +1,184 @@
...
+    this.mapData$.next(d);
+  }
+  get colorMode() {
+    return this._colorMode;
+  }
+
+  legends = legends;
+
+  private mapData$ = new BehaviorSubject<MapData>(undefined);
+  private map: any;
```
aorin, line 133: This could look a little bit cleared if there was a clear order (like all private variables are next to each other). Also, we usually place the functions last or after the constructor as it makes easier to see the variables at a glance if the functions and variables are not mixed.

aorin, line 133: This could look a little bit cleared if there was a clear order (like all private variables are next to each other). Also, we usually place the functions last or after the constructor as it makes easier to see the variables at a glance if the functions and variables are not mixed.

aorin (changes requested):

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
Blodir: I agree, but the issue with moving the atlasGrid setter is that then the inputs (and outputs) of the component are not at the same place... I made some changes:
* both map and table: removed setter from `atlasGrid` and moved the setter logic to `ngOnChanges`
* both map and table : reorganized the class variables to the following order:
  * `@Input`
  * `@Output`
  * template variables etc.
  * public variables
  * private variables
* table: moved cols intialization to the constructor
* table: proper typing for datatable rows
* map: moved colorMode setter logic to a private method so that the setter can be one lined
* map: refactored activity category colors so that the logic is equivalent to species count colors
* map: inlined precomputed gradients as was mentioned in original PR description. There's no reason to re-compute the gradients in runtime. Btw I do understand that writing the gradient logic was unnecessary and we could've just used some third party tool to pre-compute the gradients, but it was fun so I wanted to do it 😄

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
```diff
@@ -0,0 +1,174 @@
...
+};
+
+// Performance optimization: precomputing gradients
+// import { discreteColorGradient } from './color-math';
+// const acGradient = discreteColorGradient('FAFAD1', '4B57A4', 6);
+// const scGradient = discreteColorGradient('FAFAD1', '4B57A4', 4);
+const acGradient = ['fafad1', 'cbf2ad', '8ee59a', '72d5b9', '59a3c2', '4b57a4'];
+const scGradient = ['fafad1', '9bea98', '69cfc6', '4b57a4'];
+
+const getAtlasActivityCategoryColor = (ac: string): string => (acGradient[ac]);
```
aorin, line 26: I'm little confused about this change... Wasn't `ac` supposed to be something like `MY.atlasActivityCategoryEnum0` so how can you use it to access array? I can see that you have changed the format of the test data that is generated in this component but how about the production data?

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
aorin: I think the component looks very good and clear now :smiley: It doesn't matter so much where the functions are if they are one-lined/short

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
Blodir: Yep, accidentally changed the format of the test data and then compensated by changing the logic. Fixed.

projects/bird-atlas/src/app/+grid/grid-index/grid-index-map/grid-index-map.component.ts
```diff
@@ -9,6 +9,13 @@ import { BehaviorSubject, Subject } from 'rxjs';
 import { filter, takeUntil } from 'rxjs/operators';
 import { AtlasGrid, AtlasGridSquare } from '../../../core/atlas-api.service';
 
+type AtlasActivityCategory =
```
aorin, line 12: This type definition could be on the `atlas-api.service` where the other types are (so then the variable `activityCategory` of the `AtlasGridSquare` interface would be like this `activityCategory?: {key: AtlasActivityCategory; value: string}`)

aorin (approved):
:+1:

PR #288, opened 2022-05-06
Fix annotation system on pap 180175905 vir

MLumme:
Fixed this problem: https://www.pivotaltracker.com/story/show/180175905, testable at https://180175905.viranomaiset-dev.laji.fi/


olzraiti (approved):
:+1:

PR #292, opened 2022-05-13
More more more more strict typing [#180206625]

olzraiti:
https://220506.dev.laji.fi/

### DIFF
--- a/projects/laji/src/app/shared/service/browser.service.ts
+++ b/projects/laji/src/app/shared/service/browser.service.ts
@@ -107,7 +107,7 @@ export class BrowserService implements OnDestroy {
       visibilityChange = 'webkitvisibilitychange' as keyof DocumentEventMap;
     }
 
-    if (!visibilityChange) {
+    if (!visibilityChange || !hidden) {
       return;
     }
 
--- a/projects/laji/src/app/shared/service/browser.service.ts
+++ b/projects/laji/src/app/shared/service/browser.service.ts
@@ -29,10 +29,10 @@ export class BrowserService implements OnDestroy {
   lgScreen$ = this.state$.pipe(map((state) => state.lgScreen), distinctUntilChanged());
   visibility$ = this.state$.pipe(map((state) => state.visibility), distinctUntilChanged());
 
-  private resizingSub: Subscription;
-  private resizeSub: Subscription;
-  private visibilityChangeEvent: string;
-  private handlerForVisibilityChange: (doc: Document) => any;
+  private resizingSub?: Subscription;
+  private resizeSub?: Subscription;
+  private visibilityChangeEvent?: string;
+  private handlerForVisibilityChange?: EventListener;
   private resize = new Subject();
 
   constructor(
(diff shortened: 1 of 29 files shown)


projects/laji/src/app/shared/service/confirm-modal.component.ts
```diff
@@ -27,16 +27,16 @@ import { BsModalRef } from 'ngx-bootstrap/modal';
...
-  message: string;
+  message!: string;
   confirmLabel = 'OK';
   cancelLabel = 'cancel';
   prompt = false;
   promptValue = '';
   showCancel = true;
 
-  value: boolean | string | null;
+  value!: boolean | string | null;
```
aorin, line 37: I'm wondering what does the exclamation mark means here? I thought that it means that we tell the compiler that it can't be null but then its type has null?

aorin, line 37: I'm wondering what does the exclamation mark means here? I thought that it means that we tell the compiler that it can't be null but then its type has null?

aorin (changes requested):
Thanks for working on this, looks good. Some minor style comments below (that seem to duplicated for some reason...)

projects/laji/src/app/shared/service/browser.service.ts
```diff
@@ -90,24 +90,31 @@ export class BrowserService implements OnDestroy {
...
+    if (!visibilityChange) {
+      return;
     }
 
     this.handlerForVisibilityChange = () => {
+      if (!hidden) {
+        return;
+      }
       this.updateState({..._state, visibility: !this.document[hidden]});
     };
```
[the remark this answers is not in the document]
olzraiti: Yeah true, I made it that way. I had to keep the `!hidden` check in the event handler though, as the TS compiler can't be sure that the value is defined at line 118 in strict mode.

projects/laji/src/app/shared/service/confirm-modal.component.ts
olzraiti: Yeah that's what it does, it's the non-null assertion operator. It's indeed a bit confusing, but adding it makes the compiler shut up about this:

```
Error: projects/laji/src/app/shared/service/confirm-modal.component.ts:37:3 - error TS2564: Property 'value' has no initializer and is not definitely assigned in the constructor.

37   value: boolean | string | null;
 ```

So in this context, it kind of makes sense that we are telling the compiler that `value` will have a sane value (and we take responsibility of it by adding the non-nullish assertion) even though it's not initialized at declaration. It's similar to `any`: it makes the compiler to skip the type checkings and you take responsibility of the integrity of the typings. `any` and `!` could be considered antipatterns, but our goal is to just get the strict mode on, so it's justifiable here.

aorin: Ok, makes sense, thanks for the explanation!

aorin (approved):

PR #293, opened 2022-05-17
Fix observation-form taxon list changes are handled in separate method [#182206341]

olzraiti:
Fixes this bug: https://www.pivotaltracker.com/story/show/182206341

The problem was that the `observation-form` taxon pill list changes were handled with `updateSearchQuery()` method, which is used by many other parts of the component to update the search query. The bug surfaced only now because `updateSearchQuery()` has before the polygon search feature received only `string` values (now it receives a `number` from the coordinates intersection buttons, and the `value.indexOf` check fails.


MLumme (approved):
Looks fine, no problems in testing.

PR #294, opened 2022-05-18
Fix document-form annotation cache

olzraiti:
Document form failed for documents that had so many annotations that they resulted in multiple paged requests. I added the page to the annotation cache key, so it works.

This is hard to debug, but I think the error is so obvious that a simple code review is enough. I tested that it works.

### DIFF
--- a/projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
+++ b/projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
@@ -509,11 +509,11 @@ export class DocumentFormFacade {
   }
 
   private getAnnotations(documentID: string, page = 1, results = []): Observable<Annotation[]> {
-    if (this.annotationCache[documentID]) {
-      return this.annotationCache[documentID];
+    if (this.annotationCache[this.getAnnotationCacheKey(documentID, page)]) {
+      return this.annotationCache[this.getAnnotationCacheKey(documentID, page)];
     }
 
-    this.annotationCache[documentID] = (!documentID || FormService.isTmpId(documentID))
+    this.annotationCache[this.getAnnotationCacheKey(documentID, page)] = (!documentID || FormService.isTmpId(documentID))
     ?  of([])
     : this.lajiApi.getList(
       LajiApi.Endpoints.annotations,
@@ -526,6 +526,10 @@ export class DocumentFormFacade {
       shareReplay(),
     );
 
-    return this.annotationCache[documentID];
+    return this.annotationCache[this.getAnnotationCacheKey(documentID, page)];
+  }
+
+  private getAnnotationCacheKey(documentID: string, page: number) {
+    return documentID + page;
   }
 }
--- a/projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
+++ b/projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
@@ -509,11 +509,12 @@ export class DocumentFormFacade {
   }
 
   private getAnnotations(documentID: string, page = 1, results = []): Observable<Annotation[]> {
-    if (this.annotationCache[this.getAnnotationCacheKey(documentID, page)]) {
-      return this.annotationCache[this.getAnnotationCacheKey(documentID, page)];
+    const cacheKey = documentID + page;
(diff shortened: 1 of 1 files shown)


projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
```diff
@@ -509,11 +509,11 @@ export class DocumentFormFacade {
   }
 
   private getAnnotations(documentID: string, page = 1, results = []): Observable<Annotation[]> {
-    if (this.annotationCache[documentID]) {
-      return this.annotationCache[documentID];
+    if (this.annotationCache[this.getAnnotationCacheKey(documentID, page)]) {
```
aorin, line 512: This long line is repeated so many times that I think it could be assigned to variable `const cacheKey = this.getAnnotationCacheKey(documentID, page)`

projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
```diff
@@ -526,6 +526,10 @@ export class DocumentFormFacade {
       shareReplay(),
     );
 
-    return this.annotationCache[documentID];
+    return this.annotationCache[this.getAnnotationCacheKey(documentID, page)];
+  }
+
+  private getAnnotationCacheKey(documentID: string, page: number) {
```
aorin, line 532: Return type could be added

aorin (changes requested):
Looks good but the style could be little improved

projects/laji/src/app/+project-form/form/document-form/document-form.facade.ts
aorin: Now that I looked it more, the cache seems to work little weird. When page = 1, it saves all the whole annotation list to cache, when page = 2, it saves all the annotations from page 2 onwards to cache and so on... So the cache seems unnecessary if the page is something else than 1

aorin: One suggestion for fixing this is to rename `getAnnotations` to `getAnnotationsFromPage` and make a new `getAnnotations(documentID)` function that handles the caching and calls `getAnnotationsFromPage(documentID, 1)`

olzraiti: I removed the `this.getAnnotationCacheKey` altogether, since the cache key can be created in a single line.

> Now that I looked it more, the cache seems to work little weird. When page = 1, it saves the whole annotation list to cache, when page = 2, it saves all the annotations from page 2 onwards to cache and so on... So the cache seems unnecessary if the page is something else than 1

Hmm... I don't find the cache unnecessary. The `getAnnotations()` works recursively, so it will first get the first page, and then get the consecutive pages if needed and all the pages are cached.

aorin: Yeah it works recursively so the the cache will look like this
`annotationCache[doc + 1] = Observable<annotations from page 1 to 100>`
`annotationCache[doc + 2] = Observable<annotations from page 2 to 100>`
`annotationCache[doc + 3] = Observable<annotations from page 3 to 100>`
`...`
So in this scenario it will cache annotations that are in page 100 100 times...

olzraiti: > So in this scenario it will cache annotations that are in page 100 100 times...

I think it fetches the page 100 only once. I see it like this:

```
getAnnotations(doc, 1) = annotationCache[doc, 1] + Observable<annotations from page 1> + getAnnotations(doc, 2) 
getAnnotations(doc, 2) = annotationCache[doc, 2] + Observable<annotations from page 1> + getAnnotations(doc, 3) 
...
getAnnotations(doc, 99) = annotationCache[doc, 99] + Observable<annotations from page 1> + getAnnotations(doc, 100) 
getAnnotations(doc, 100) = annotationCache[doc, 100]
```

The results are stored to memory in each call though (`[...results, ...result.results]`) before the call stack is complete thought, which isn't optimal... Realistically that isn't a problem. I wonder if we have some page iteration utility somewhere for cases like this.

aorin (approved):

PR #299, opened 2022-05-31
Annotate Soundscapes section to Bird Sounds Global

aorin:
This pull request adds a new section, Annotate Soundscapes, to the bsg site. All the old content has been moved under Validate Species Templates section. Can be tested here: https://kerttu-global-dev.laji.fi/identification

This is a huge pull request but maybe it's not necessary to go through all the files in a very detailed way since most of the changes don't affect other places. Also, there is no hurry for this to get accepted.

The code related to the new section is under the folder
`projects/kerttu-global/src/app/+identification`
All the files in that folder are new files

The code related to the old section has been moved under this folder
`projects/kerttu-global/src/app/+validation`
There shouldn't be any major changes

This pull request has some changes to shared components which are listed here and should be checked in more detail:

- `projects/laji-ui/src/lib/mixins/_base.scss` 
    The class `.laji-page` made some parts of the `typeahead` component and `laji-map` look weird so I fixed those
- `projects/laji/src/app/shared-modules/ykj/ykj-map/ykj-map.component.css`
   `projects/laji/src/styles/laji.scss`
   I moved ykj map legend style classes to shared styles
- `projects/laji/src/app/shared/panel/panel.component.html`
   `projects/laji/src/app/shared/panel/panel.component.ts`
   I added an ability to give a custom panel header template to the panel component
- `projects/laji/src/app/shared/model/Area.ts`
  `projects/laji/src/app/shared/service/area.service.ts`
   I added a new area type, continent
- `projects/laji/src/app/shared/model/Profile.ts`
   I added new profile fields


projects/laji/src/app/shared/navbar/navbar.component.html
```diff
@@ -45,7 +45,7 @@
                 <a [routerLink]="['/validation'] | localize">{{ 'validation.title' | translate }}</a>
               </li>
               <li [routerLinkActive]="['active']">
-                <a [routerLink]="['/results'] | localize">{{ 'results.title' | translate }}</a>
+                <a [routerLink]="['/identification'] | localize">{{ 'identification.title' | translate }}</a>
```
Blodir, line 48: This is unrelated to this specific PR, but as a sidenote imo in the future we should avoid cluttering laji project components with stuff that is related to other projects. It would be much easier to simply have a new navbar component for each project.

projects/laji/src/styles/laji.scss
```diff
@@ -1360,3 +1360,41 @@ span.simple-option {
     min-width: 250px !important;
   }
 }
+
+.map-legend {
```
Blodir, line 1364: Could this be moved to a new file? Eg. `src/styles/map/ykj-map.scss`. Also the map-legend name is maybe a bit too general for global namespace so either it should be renamed, or imported only where it's needed.

projects/laji-ui/src/lib/mixins/_base.scss
```diff
@@ -209,4 +209,45 @@
     margin: 20px 0;
     border-top: 1px solid $neutral-4;
   }
-}
\ No newline at end of file
+
+  typeahead-container, .laji-map {
```
Blodir, line 213: I would prefer if the base styles would remain decoupled from laji-map and bootstrap. The laji-map and bootstrap related stuff could be moved eg. to `laji.scss` where the base styles are `@include`d.

Blodir (changes requested):

projects/laji/src/styles/laji.scss
aorin: I renamed the class as map-color-legend and moved it to its own file

projects/laji-ui/src/lib/mixins/_base.scss
aorin: Moved there

projects/laji/src/app/shared/navbar/navbar.component.html
aorin: I agree. I copied this style from the iucn project but both of these projects could have their own navbar component. It needs some refactoring (as the general app component includes several features such as login, language selection and feedback) so maybe it could be its own task

Blodir (approved):

PR #303, opened 2022-05-31
Observation search filter has dropped out for some reason

MLumme:
"Kansalliset kokoelmat"-filter was missing a query parameter, task: https://www.pivotaltracker.com/story/show/181768674, branch: https://181768674.dev.laji.fi/observation/list.


olzraiti (approved):
LGTM

PR #305, opened 2022-06-02
Kerttu closing

aorin:
Testable here: https://182331968.dev.laji.fi/theme/kerttu

Replaces Kerttu pages with an info page about the closing. I also removed all the Kerttu related imports from the `theme.module.ts`


olzraiti (changes requested):
The kerttu related components could also be removed, if they aren't used anymore.

aorin:
I wouldn't remove them as there will be similar project in the future and they could be useful then. They could be moved somewhere but I'm not sure where

olzraiti (approved):
Thanks!

PR #321, opened 2022-06-29
Show GIS download progress

aorin:
Shows a progress percent when downloading a GIS file
Can be tested here: https://182008442.dev.laji.fi/citation/HBF.8797


projects/laji/src/app/+observation/download/observation-download.component.ts
```diff
@@ -74,6 +74,7 @@ export class ObservationDownloadComponent implements OnDestroy {
   requests: {[place: string]: RequestStatus} = {};
   requestStatus = RequestStatus;
   downloadLoading = false;
+  downloadProgressPercent?: number;
```
olzraiti, line 77: I wonder if the logic of `downloadLoading` and `downloadProgressPercent` could be combined into one single variable by just removing `downloadLoading` and then a numeric value of `downloadProgressPercent` would imply that a download is loading. A single source of truth helps preventing bugs in the future and simplifies things.

Is there some specific reason for using `null` for `downloadProgressPercent`, instead of `undefined`? Now the value for `downloadProgressPercent` can be either `undefined` (the initial value), `null` or a numeric value. We should aim at using only either `undefined` or `null` for empty values in order to keep things consistent. We probably use `undefined` more often, and that's probably for the best since it eliminates cases where there are multiple possible values for nil values (non-initialized values are `undefined` anyway, so the solutions that make sense are either 1) use `undefined` or 2) use `null`, and initialize the variable with `null`. The 1st option is less work.)

projects/laji/src/app/shared-modules/download-modal/download.component.ts
```diff
@@ -96,6 +101,7 @@ export class DownloadComponent implements OnChanges {
 
   @Input() disabled = false;
   @Input() downloadLoading = false;
+  @Input() progressPercent?: number;
```
olzraiti, line 104: Same case as in the review message above

projects/laji/src/app/shared/service/geo-convert.service.ts
```diff
@@ -92,17 +102,20 @@ export class GeoConvertService {
...
     return interval(this.pollInterval).pipe(
-      mergeMap(() => this.getGeoConversionStatus(conversionId)),
-      first(result => result.status === 'complete'),
-      map(() => '/api/geo-convert/output/' + conversionId)
+      concatMap(() => this.getGeoConversionStatus(conversionId)),
+      takeWhile(result => result.status !== 'complete', true),
+      map((result) => ({
+        status: result.status,
+        progressPercent: result.progress_percent,
+        outputLink: result.status === 'complete' ? '/api/geo-convert/output/' + conversionId : null
```
olzraiti, line 112: Same reasoning for `undefined` vs `null`, `undefined` would be better if there isn't some specific reason for using `null`. The compiler will show an error for the typing of `outputLink` for `null` value once we enable `strictNullChecks` for TS.

olzraiti (changes requested):

projects/laji/src/app/+observation/download/observation-download.component.ts
aorin: The variables `downloadLoading` and `downloadProgressPercent` are separate because only the GIS download has the progress percent and we don't want to show any percent if the percent is not available.

I replaced those `null` values with `undefined`

olzraiti: Thanks! I added a few commits that made the `laji-download` & `laji-file-download` html templates to check for `undefined` instead of `null`. If it looks OK to you we can approve this

aorin: I changed those to strict equalities (`!==`) since the variable can't be anything else than a number or undefined

olzraiti: Good catch, thanks!

olzraiti (approved):

PR #323, opened 2022-07-27
Make filter subtitle localizable

wkmor1:
Currently can't change the filter section header "Administrative status"
in Crowdin. This fix should make that possible.


MLumme (approved):
Worked as expected when tested.

PR #325, opened 2022-08-18
Fix vihko import not parsing coordinates that contain spaces

aorin:
https://www.pivotaltracker.com/story/show/182713382


olzraiti (approved):
LGTM!

PR #326, opened 2022-08-18
Notification queries 180080394

Blodir:
https://www.pivotaltracker.com/story/show/180080394

Fixes (most instances of) double notification queries and changes the time interval to 5min. From pivotal:

>Updating unseen notifications is done through regular graphql queries (now at 5 minute intervals), but once the notifications dropdown is opened the list (which uses virtual scrolling) uses the regular api.


aorin (approved):
Looks good, good that you added comments since it's quite complex system

PR #332, opened 2022-08-26
Image license links

Blodir:
https://www.pivotaltracker.com/story/show/182700426

Also refactored image overlay a bit:
- Got rid of `childComunicationService`
- Also got rid of magic empty class method `close`
- Replaced with explicit output stream for the modal show/close state
- There's a constant stream of change detection cycles (I'm not sure what causes this still, but it's not good) so I made the component OnPush to improve performance


olzraiti (approved):
nice!

PR #334, opened 2022-09-09
Patch for v58: reactive observation map fixes[#183164851]

olzraiti:
https://www.pivotaltracker.com/story/show/183164851
https://183164851.dev.laji.fi/observation/map

Fixes the following bugs:
* doesn't show individual points for finnish layers
* when results are filtered so that there are <10K results, the map shouldn't make further queries on zoom/pan

There is still at least one bug (documented in the pivotal task discussion), but we left it out of scope for this patch. The component will probably need a rewrite the next time it's going to get bigger changes...


projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -462,8 +471,22 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
       );
   }
 
-  private addViewPortCoordinates(query: WarehouseQueryInterface) {
-    if (!query.coordinates && this.activeZoomThresholdBounds && this.activeZoomThresholdLevel >= this.onlyViewportThresholdLevel) {
+  private queryInsideViewport(query: WarehouseQueryInterface) {
```
Blodir, line 474: return type boolean? Also I guess `queryIsInsideViewport` would be more correct

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -554,13 +572,20 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
+      this.drawDataSubscription = this.getDrawFeatureCollection().subscribe(featureCollection => {
         this.drawData = {...this.drawData, featureCollection};
         mapData.push(this.drawData);
         this.mapData = mapData;
       });
     }
   }
+
+  private updateTotalCount() {
+    this.getCountForQuery$(this.query).subscribe(count => {
```
Blodir, line 584: Requires unsubscription? I wonder actually, if the query takes a while and the component is destroyed before it completes, does this produce a `this === undefined` error or not (since `this` might not be actually gone/garbage collected yet)

Blodir (changes requested):
At least the unsubscription thing should be checked, but after that I think it's ready 👍

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
olzraiti: Interesting question... I don't think I've ran into `this` being `undefined` on an angular component ever - I'm not sure if the RXJS subscription would somehow keep the instance alive? I added an unsubscription anyways, so this isn't even a theoretical problem anymore :)

Blodir (approved):

PR #335, opened 2022-09-15
Use JSON Schema oneOf schematic instead of enum & enumNames [#183287493]

olzraiti:
https://183287493.dev.laji.fi/
https://www.pivotaltracker.com/story/show/183287493

Form backend received changes so that the schema format has changed form enums like this:

`enum: ["MX.a", "MX.b"], enumNames: ["aLabel", "bLabel"]`
will be like this:
`oneOf [{const: "MX.a", title: "aLabel"}, {const: "MX.b", title: "bLabel"}]`

Reasoning (copy-pasted from pivotal): 

> Current form backend serves alt ranges following a rejected JSON Schema proposal [https://github.com/json-schema-org/json-schema-spec/issues/57](which got rejected). In order to serve a proper standard, we are migrating from the `enum` & `enumNames` schema to using `oneOf` (in format `{const: string, title: string}`.


esko-piirainen:
@aorin Would be good to check if this breaks 1) label designed 2) vihko import

projects/laji/src/app/shared/model/Form.ts
```diff
@@ -37,10 +37,7 @@ export namespace Form {
     options: ListOptions;
   }
 
-  export interface IEnum {
-    enum: string[];
-    enumNames: string[];
-  }
+  export type IEnum = {const: string; title: string}[];
```
aorin, line 40: Should this interface be `oneOf: {const: string; title: string}[];` (like in the file `projects/laji-api-client/src/lib/models/UISchemaContext.ts` above)?

aorin (commented):
Looks good, I tested the Vihko import and it seems to work. Because there are changes to label designer, we must remember to update the label designer npm package before the update (can't do it before merging to development since there are other pull requests with changes to label designer).

projects/laji/src/app/shared/model/Form.ts
olzraiti: > Because there are changes to label designer, we must remember to update the label designer npm package before the update

Right. The label designer schema service is used only on laji.fi, not on Kotka right? Does laji.fi use the label-designer from the npm package, or from directly from the source? If it uses the label designer from npm, we have to make a release for it and include a dependency update to this PR... It's in package.json deps at least so I assume we have to do it that way. So this PR is recursively depending on it's changes. Pretty crazy stuff!

aorin: That's correct but [this pull request](https://github.com/luomus/laji/pull/337) needs to be merged first because I made release 4.0.6 for label designer there

aorin (approved):

projects/laji/src/app/shared/model/Form.ts
olzraiti: Ah, so I'll merge that to dev now and then merge dev here.

PR #336, opened 2022-09-28
Vihko import fix

aorin:
https://www.pivotaltracker.com/story/show/182750980
https://www.pivotaltracker.com/story/show/183101256

Fixes Vihko import failing silently with data that is missing document/gathering/unit level data. This pull request also changes the order of rows in the summary table so that the erroneous rows are shown first.


MLumme (approved):
Looks ok, approved.

PR #345, opened 2022-11-04
Changed PAP navbar title. [#183697660]

rpulkka:
Now the title in PAP has a translation, which is shown in uppercase letters. Fixes 2022.

### DIFF
--- a/projects/vir/i18n/fi.json
+++ b/projects/vir/i18n/fi.json
@@ -5,6 +5,7 @@
   "navigation.observation": "Havaintotieto",
   "navigation.saveVirObservations": "Tallenna havaintoja",
   "navigation.taxonomy": "Lajitieto",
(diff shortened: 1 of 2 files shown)


olzraiti (changes requested):
Hmm... It's kind of more dynamic now that it uses `toUpperCase()`, but on the other hand I think having simply `"VIRANOMAISPORTAALI"` as the translation and not do runtime-uppercasing would be better, since it's more searchable in the code base, and it would be more controllable (if we decide to have it in lower case in the future, it can be updated in Crowdin by anybody)

rpulkka:
You're right, having the uppercase styling in the translation sounds more practical. I changed the code accordingly.

olzraiti (approved):
:+1: Thanks!

PR #346, opened 2022-11-04
Missing dev-ribbon added to PAP. [#183711188]

rpulkka:
Not sure how to check if this works on production. Caught a bug where dev-ribbon was not shown on PAP's dev website. This was because the environment variables on PAP, such as environment.production function unintuitively, so the dev-ribbon display logic did not work as expected. A new env-variable was added to check whether to display the dev-ribbon on PAP, while applying the previous logic to other services in the monorepo.

### DIFF
--- a/projects/laji/src/app/shared/navbar/navbar.component.ts
+++ b/projects/laji/src/app/shared/navbar/navbar.component.ts
@@ -10,6 +10,7 @@ import {
 import { UserService } from '../service/user.service';
 import { NavigationEnd, Router } from '@angular/router';
 import { environment } from '../../../environments/environment';
+import { environment as virEnvironment } from '../../../../../vir/src/environments/environment';
 import { LocalizeRouterService } from '../../locale/localize-router.service';
 import { TranslateService } from '@ngx-translate/core';
 import { timer, Subject, Observable } from 'rxjs';
@@ -54,7 +55,9 @@ export class NavbarComponent implements OnInit, OnDestroy {
     private ngZone: NgZone
   ) {
     this.navId = environment.type + '-nav';
-    this.devRibbon = !environment.production || environment.type === Global.type.beta;
+    this.devRibbon = environment.type === Global.type.vir ?
+      virEnvironment.displayDevRibbon :
+      !environment.production || environment.type === Global.type.beta;
     this.redTheme = environment.type === Global.type.vir || environment.type === Global.type.iucn;
     this.containerClass = environment.type === Global.type.iucn ? 'container' : 'container-fluid';
   }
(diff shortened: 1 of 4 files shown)


projects/laji/src/app/shared/navbar/navbar.component.ts
```diff
@@ -54,7 +55,9 @@ export class NavbarComponent implements OnInit, OnDestroy {
     private ngZone: NgZone
   ) {
     this.navId = environment.type + '-nav';
-    this.devRibbon = !environment.production || environment.type === Global.type.beta;
+    this.devRibbon = environment.type === Global.type.vir ?
+      virEnvironment.displayDevRibbon :
+      !environment.production || environment.type === Global.type.beta;
```
olzraiti, line 60: I think the `displayDevRibbon` should have highest priority since it's explicitly stating that the ribbon should be shown. Now the flag is checked for `vir` environment. Also, you're not supposed to check the `vir` project's environment file directly. The environment should be always imported from `projects/laji/src/environments/environment.ts` in code. The build processes handle rewriting that file (see `angular.json`, it contains the file replacement logic).

Proposal:

1. Set `devRibbon` for beta env, so the `environment.type === Global.type.beta` check can be removed.
2. Change the logic to:
```
this.devRibbon = environment.displayDevRibbon ||  environment.production === false
```

As a result, the `displayDevRibbon` is shown when explicitly stated in the environment. Checking a falsy `environment.production` is also OK, because it should never be `false` in production environment.

olzraiti (changes requested):

projects/laji/src/app/shared/navbar/navbar.component.ts
rpulkka: Thanks for the feedback! I have one question: How to avoid warnings when referring to "environment.displayDevRibbon" in step 2 of the proposal? TypeScript does not seem to be aware that the content of environment.ts gets rewritten, it just sees that displayDevRibbon cannot be found in that file directly.

olzraiti: That can be fixed by adding typings for the base environment (`projects/laji/src/environments/environment.ts`), or adding `displayDevRibbon` to the base environment properties so TS can infer the type correctly. The latter option would be less work? I think `displayDevRibbon: false` would make sense since it's supposed to be false by default.

olzraiti (approved):
:+1: LGTM

rpulkka:
Re-requested the review, as I found out that adding the new variable in the default file was deleted when the specified env-files overwrote it. Now the variable is added to all specified env-files, I hope I did not forget any of them.

olzraiti (approved):
:+1:

PR #348, opened 2022-11-09
exhibition screens: confusing type error fix

Blodir:
https://www.pivotaltracker.com/story/show/183125847


olzraiti (approved):
I think we'll run into this issue in the future. It makes no sense that `removeUndefinedFromObject` doesn't allow the `Lang` interface here... I recall running into this on some other project, and in that project I fixed this by replacing all interfaces with types. I tested if that would fix that, and indeed if I change `Query.Lang` to be a type instead of an interface, the error goes away.

Well, fixing this "deeply" is out of scope. Let's just get on with it - approved.

PR #349, opened 2022-11-09
Fix multiple tooltips 183069115

Blodir:
https://www.pivotaltracker.com/story/show/183069115

Note: test that there's not info components inside buttons (if there is, this should not be merged)


MLumme (changes requested):
Otherwise fine, but I found one instance of laji-info inside lu-button, np-info.component.html, line 33, which you might want to check.

MLumme (approved):
Looks fine now, approved.

PR #350, opened 2022-11-09
Always show selected dataset in hierarchy 183624947

MLumme:
If page is loaded with an empty collection selected the tree switches to showing empty collections and selects it from the hierarchy, opening the tree to its level. Testable branch https://183624947.dev.laji.fi/theme/dataset-metadata/.


rpulkka (approved):
Looks good, approved. We discussed that some functional additions may be coming up next sprint.

PR #352, opened 2022-11-09
Checkbox for showing hidden taxa [#183710973]

rpulkka:
https://183710973.dev.laji.fi/taxon/MX.70285?showTree=true&showHidden=true
https://www.pivotaltracker.com/n/projects/2346653/stories/183710973

The new checkbox shows hidden taxa in the taxonomy tree. There's also an URL parameter, that makes it possible to send a link, where the checkbox is selected. Part of the Fish Checklists -epic.


MLumme (approved):
Looks fine, approving.

PR #353, opened 2022-11-11
Kerttu global changes

aorin:
Can be tested here: https://kerttu-global-dev.laji.fi/

Several changes to Kerttu global project. They have been in prod quite a long now, I had forgotten to do this PR... This PR doesn't have many changes to shared stuff, here are the main ones:

- `projects/laji/src/app/shared-modules/info/info/info.component.html`
  - fine-tuned info box style when `showOnHover` is true (I don't think this option is used anywhere else than Kerttu global)
- `projects/laji/src/app/shared/navbar/navbar.component.html`
  - removed Kerttu global related things


MLumme (approved):
Looks good, while huge I did try to check it trough but didn't notice any obvious problems, plus you'd think problems would be noticeable in use, so approved.

PR #354, opened 2022-11-18
BOLD User interface [#183675507]

rpulkka:
https://183675507.dev.laji.fi/taxon/MX.60841?showTree=true
https://www.pivotaltracker.com/n/projects/2346653/stories/183675507

BOLD links can now be found on the overview- and taxonomy-tabs of taxa, under the DNA-barcode sequences -section.


projects/laji/src/app/+taxonomy/taxon/info-card/shared/component/taxon-bold/taxon-bold.component.scss
```diff
@@ -0,0 +1,3 @@
+.bold-entry {
```
aorin, line 1: Just a note, we have utility classes for margins (for example `mb-1` gives small bottom margin) so those could be used instead of creating new classes

aorin (approved):
Looks good

projects/laji/src/app/+taxonomy/taxon/info-card/shared/component/taxon-bold/taxon-bold.component.scss
rpulkka: Thanks for the tip, tried to utilize mb-1 and to remove the unnecessary new styles now.

PR #357, opened 2022-11-28
BOLD Filter for the taxon list [#183674896]

rpulkka:
https://183674896.dev.laji.fi/taxon/list?onlyFinnish=true&hasBoldData=true
https://www.pivotaltracker.com/n/projects/2346653/stories/183674896

Added a BOLD-filter for the taxon list. There's a checkbox for showing only taxa with hasBold=true, another one for showing only taxa with hasBold=false, and empty selection shows both kinds of taxa.


projects/laji/src/app/+taxonomy/species/species-form/species-form.component.ts
```diff
@@ -137,6 +160,20 @@ export class SpeciesFormComponent implements OnInit, OnDestroy {
     this.invasiveSelected = invasiveSelected;
   }
 
+  private updateBoldSelected() {
+    const boldSelected: (keyof SpeciesFormQuery)[] = [];
```
aorin, line 164: The types used here are making this little hard to read. It looks like the value can only be either `onlyBold` or `onlyNonBold` so it doesn't make sense to use `(keyof SpeciesFormQuery)` or `(keyof Omit<SpeciesFormQuery, 'taxon'>)` (it's even more confusing that the types are different). I would declare a new type on top of the file like this `type BoldField = keyof (Pick<SpeciesFormQuery, 'onlyBold' | 'onlyNonBold'>)` and then use it in all places (`public boldSelected: BoldField[] = [];`)

aorin (changes requested):

projects/laji/src/app/+taxonomy/species/species-form/species-form.component.ts
rpulkka: Good point! I had tried to copy what some other filters were using, but did not realize there was a smarter way to do it. I tried to apply the suggested changes now, and requested a new review.

aorin (approved):
Thanks, looks good!

PR #359, opened 2022-12-13
Fix GIS download selection resetting

aorin:
https://www.pivotaltracker.com/story/show/183838261

If the user filled the GIS download form before the page had fully loaded, it reset. This was fixed by disabling radio buttons and lu-buttons when server side rendering.


Blodir (approved):
Looks good to me

PR #360, opened 2022-12-13
Fix label designer pdf issues

aorin:
https://www.pivotaltracker.com/story/show/183943142

This changes how the label designer QR codes are scaled because the previous method (css transform scale) caused issues when printing.


Blodir (approved):
Looks good to me

PR #361, opened 2022-12-13
Allow more taxon name types in vihko taxon autocomplete

aorin:


rpulkka (approved):
Approving, code changes match what is described on Pivotal.

Could not test the feature myself, because I have no valid files for the importer, but this isn't an issue as it has been approved by the product owner on Pivotal.

rpulkka:
Actually I got a test file, so now I can also confirm that this functions correctly :)

PR #362, opened 2022-12-13
Fix ykj map having wrong default label

aorin:
Can be tested here: https://183945195.dev.laji.fi/project/MHL.3/stats?tab=species&species=MX.26277&year=2022

https://www.pivotaltracker.com/story/show/183945195


rpulkka (approved):
Code changes looking good!

Note for next time: It took me a while to understand what the feature was about since there was no description where to look for it, in Pivotal or the pull request, but I managed to find it eventually.

PR #363, opened 2022-12-14
Occurrence visualization

Blodir:
https://www.pivotaltracker.com/story/show/183185605
https://www.pivotaltracker.com/story/show/183186103
https://www.pivotaltracker.com/story/show/183185809
https://www.pivotaltracker.com/story/show/183220403

https://183185605.dev.laji.fi/observation/map

Notes:
* I'd like to hear opinion(s) about whether the legend should be moved up in hierarchy (so that it is not a child of laji-map, but laji-observation-map, or perhaps its own module?). I'm leaning on moving it to its own module and instanciating it in observation-map.


projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
```diff
@@ -0,0 +1,299 @@
...
+const sliceYear = (newestRecord: string): number => (
+  parseInt((newestRecord).slice(0, 4), 10)
+);
+
+const getRecordAgeColor = (newestRecords: string | string[]): string => {
+  if (!newestRecords) { return '#ffffff'; }
+  if (!Array.isArray(newestRecords)) { return lajiMapObservationVisualization.recordAge.categories[getRecordAgeColorIdx(sliceYear(newestRecords))].color; }
+  const a = newestRecords.map(s => sliceYear(s));
+  const year = Math.max(...a);
+  return lajiMapObservationVisualization.recordAge.categories[getRecordAgeColorIdx(year)].color;
```
olzraiti, line 113: Looks familiar. Similarly to the other color getters I commented about, this could be simplified a bit (treat input as array, use default value). Now that I've read this far, I wonder if these functions could be refactored to use some shared logic. There are maybe enough differences that it's not worth it...

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
```diff
@@ -0,0 +1,299 @@
...
+      },
+      {
+        color: '#fdbf66',
+        //label: 'Uncertain'
+        translate: 'laji-map.legend.mode.recordQuality.4'
+      },
+      {
+        color: '#f26840',
+        //label: 'Erroneous'
+        translate: 'laji-map.legend.mode.recordQuality.5'
```
olzraiti, line 196: This wouldn't need the commented labels if the translation keys would be constructed from enumerations instead of indices. But it's OK as it is.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
```diff
@@ -0,0 +1,299 @@
...
+        label: `≥ ${currentYear - yearsAgoBreakpoints[2]}`
+      },
+      {
+        color: '#fdbf66',
+        label: `≥ ${currentYear - yearsAgoBreakpoints[3]}`
+      },
+      {
+        color: '#f26840',
+        label: `< ${currentYear - yearsAgoBreakpoints[3]}`
+      }
```
olzraiti, line 291: Is the `yearsAgoBreakpoints[3]` twice on purpose here? This is why hard coded indices can produce bugs easily (first doing copy pasting and manually hard coding the changes is prone to typos, then after doing that a several times your eyes just see a repetitive pattern and don't notice that one of the indices is wrong. It's like trying to find Waldo). This could be more readable if for example there was a single data structure that would hold the year breakpoints and the colors.

olzraiti (changes requested):
In general it's good stuff. I left some change request & other comments, main concern being the hard coded indices - the data structure could be improved.

olzraiti:
> I'd like to hear opinion(s) about whether the legend should be moved up in hierarchy (so that it is not a child of laji-map, but laji-observation-map, or perhaps its own module?). I'm leaning on moving it to its own module and instanciating it in observation-map.

The legend is also used for line transect routes reservation state (https://beta.laji.fi/project/MHL.1/form), so it can't be a child of the observation map.  It's not that much code, so I don't mind it being inside the laji-map module. If you decide to move it to a separate module, that's also OK.

olzraiti:
Maintenance wise, I think the most important thing to do would be to have the logic for map colors & legend to a single-source-of-truth data structure, so that when editing the visualization only a single place would have to be edited.

projects/laji/src/app/shared-modules/laji-map/visualization/laji-map-visualization.ts
```diff
@@ -0,0 +1,13 @@
...
+export interface LajiMapVisualizationCategory { color: string; label?: string; translate?: string };
+export type LajiMapVisualization<T extends string> = Record<T,
+  {
+    label: string;
+    categories: LajiMapVisualizationCategory[];
+    getFeatureStyle?: (opt: GetFeatureStyleOptions) => PathOptions;
+    getClusterStyle?: (childCount: number, featureIdxs: number[], cluster: MarkerCluster) => PathOptions;
+    getClusterClassName?: (childCount: number, featureIdxs: number[], cluster: MarkerCluster) => string;
+  }
+>;
```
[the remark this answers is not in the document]
Blodir: Fixed, also in relation to SSR, I noticed a nasty type error while building. I'm not sure if this is a bug in angular or what, but in order to keep onVisualizationModeChange typed I had to introduce this creature `(visualizationModeChange)="onVisualizationModeChange($any($event))"` in `observation-map.component.html`

projects/laji/src/app/shared-modules/laji-map/visualization/laji-map-visualization.ts
```diff
@@ -0,0 +1,13 @@
+import { GetFeatureStyleOptions } from 'laji-map';
+import { MarkerCluster, PathOptions } from 'leaflet';
+
+export interface LajiMapVisualizationCategory { color: string; label?: string; translate?: string };
```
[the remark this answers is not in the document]
Blodir: Idk how I feel about this. Practically there probably won't be any conflicts, but just the possibility that there could be one makes me feel a bit uneasy

projects/laji/src/app/shared-modules/laji-map/visualization/legend.component.html
```diff
@@ -0,0 +1,31 @@
...
+    <h5>{{ 'laji-map.legend.coordinateAccuracy' | translate }}<laji-info>{{ 'laji-map.legend.coordinateAccuracy.info' | translate }}</laji-info></h5>
+    <ul>
+      <li class="d-flex"><div class="legend-circle accuracy-1 mr-2"></div><div class="legend-label">1 - 10 m</div></li>
+      <li class="d-flex"><div class="legend-circle accuracy-2 mr-2"></div><div class="legend-label">10 - 100 m</div></li>
+      <li class="d-flex"><div class="legend-circle accuracy-3 mr-2"></div><div class="legend-label">100 - 1000 m</div></li>
+      <li class="d-flex"><div class="legend-circle accuracy-4 mr-2"></div><div class="legend-label">1000 - 10 000 m</div></li>
+      <li class="d-flex"><div class="legend-circle accuracy-5 mr-2"></div><div class="legend-label">10 000 - 100 000 m</div></li>
+    </ul>
+  </ng-container>
+</div>
```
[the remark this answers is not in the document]
Blodir: I can't think of any way to lift the coordinate accuracy from the legend without changing the UI... This is a big reason why I'm thinking about moving the legend to its own module.

projects/laji/src/app/shared-modules/laji-map/laji-map.component.ts
```diff
@@ -30,35 +32,38 @@ import { environment } from 'projects/laji/src/environments/environment';
...
     </div>`,
   styleUrls: ['./laji-map.component.css'],
   providers: [],
   changeDetection: ChangeDetectionStrategy.OnPush
 })
 export class LajiMapComponent implements OnDestroy, OnChanges, AfterViewInit {
-
   @Input() data: any = [];
+  @Input() visualization: LajiMapVisualization<any> | undefined;
+  @Input() visualizationMode: string | undefined;
```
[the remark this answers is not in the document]
Blodir: I originally had the component as generic, but ran into some issues with that (I forgot what exactly). I don't really see an issue with the generic type as at most it adds a little bit of boilerplate when using the type. I changed the `LajiMapVisualization<any>` to `LajiMapVisualization<string>` though.

projects/laji/src/app/shared-modules/laji-map/visualization/legend.component.html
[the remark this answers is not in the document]
olzraiti: We could provide a strategy name for rendering extra stuff, that would point to a template name. Similarly to how we handle datatable cell templates. So for example, `LajiMapVisualization` could have a prop `extraTemplate`, and `lajiMapObservationVisualization` could have something like:

```
const lajiMapObservationVisualization = {
...,
extraTemplate: 'coordinateAccuracy'
```

Then there'd be a template for `coordinateAccuracy` that is used. It's a bit ugly that the actual template implementation would have to be inside the `legend.component.html`, but Angular doesn't provide any alternatives.

My initial complaint was that the observation related code is inside the the legend template code, but at least this way it'd be more extendable and the API of the legend component wouldn't have anything related to the observation stuff (other than a template strategy name).

projects/laji/src/app/shared-modules/laji-map/laji-map.component.ts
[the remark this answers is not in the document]
olzraiti: Ok, sounds reasonable

projects/laji/src/app/shared-modules/laji-map/visualization/laji-map-visualization.ts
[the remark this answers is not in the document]
olzraiti: I read the usage again, and I was confused because I didn't notice that there are two different `label` props (one for the visualization instance and then for the categories). So no problem here :)

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
```diff
@@ -0,0 +1,299 @@
...
+      return 4;
+  }
+};
+
+const getRecordQualityColor = (qualities: string | string[]): string => {
+  if (!qualities) { return lajiMapObservationVisualization.recordQuality.categories[2].color; }
+  if (!Array.isArray(qualities)) { return lajiMapObservationVisualization.recordQuality.categories[getRecordQualityColorIdx(qualities)].color; }
+  const a = qualities.map(q => getRecordQualityColorIdx(q));
+  return lajiMapObservationVisualization.recordQuality.categories[Math.min(...a)].color;
+};
```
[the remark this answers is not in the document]
Blodir: When this was first implemented with clustering the array input was required. It's no longer used so I removed it. I'm not sure how much it improves the code, but I did this with the enum:
```
enum RecordQuality { ExpertVerified, CommunityVerified, Neutral, Uncertain, Erroneous }
const recordQualityMap = {
  EXPERT_VERIFIED: RecordQuality.ExpertVerified,
  COMMUNITY_VERIFIED: RecordQuality.CommunityVerified,
  UNCERTAIN: RecordQuality.Uncertain,
  ERRONEOUS: RecordQuality.Erroneous
};
const getRecordQualityColor = (recordQuality: string | undefined): string => {
  if (!recordQuality) { return lajiMapObservationVisualization.recordQuality.categories[RecordQuality.Neutral].color; }
  return lajiMapObservationVisualization.recordQuality.categories[recordQualityMap[recordQuality]].color;
};
```

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
Blodir: `yearsAgoBreakpoints[3]` is twice on purpose.

Blodir: I made this a little clearer by moving the `currentYear - breakpoint` calculation to the `RECORD_AGE_BREAKPOINTS` array, but I'm not sure if there's anything else that can be done. The values can't be written directly into the string/enum even if we wanted to since it depends on current year (which is dynamic). Besides honestly I think the logic here is pretty clear?

olzraiti: I was wrong about the label, it's supposed to be 
`< 1982` like it is now. Sorry :D I'll mark this as resolved

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-visualization.ts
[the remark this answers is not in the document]
olzraiti: Thanks, just adding the enums makes it more readable.

Blodir:
Note that there are a few new changes since @esko-piirainen requested that instead of doing a count query we do a point query first, check if there's too many results, and do a box query if necessary. That is now implemented and testable.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -209,210 +237,272 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
-    if (outOfActiveBounds
-      || zoomedInAndNotShowingPoints
-      || zoomedOut
+    if (
+      !(this.showingIndividualPoints && activeBoundsContainViewport)
+      && (
+           // Zoom in
+           this.activeZoomThresholdLevel > curActiveZoomThresholdLevel
+           // Zoom out
+        || this.activeZoomThresholdLevel < curActiveZoomThresholdLevel
```
olzraiti, line 245: It's not a biggie, but could be shortened to `this.activeZoomThresholdLevel !== curActiveZoomThresholdLevel`

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -209,210 +237,272 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
+  resetTable() {
+    this.selectedObservationCoordinates = undefined;
+    this.cdr.detectChanges();
+    // Wait until next cycle so that height has time to adjust to the removal of the table
     setTimeout(() => {
-      this.reloading = false;
-    }, 200);
+      this.tableViewHeightOverride = undefined;
+      this.cdr.detectChanges();
+    });
```
olzraiti, line 289: Did you try could this be done at `AfterViewChecked` so no need for timeout?

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -209,210 +237,272 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
+        },
+        marker: {
+          icon: getPointIconStyle
+        },
+        total: data.total
+      }))
+    );
+  }
+
+  private onAggregateDataClick(args: any[]) {
```
olzraiti, line 363: Typing could be improved

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -209,210 +237,272 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
+          firstPage.features.push(...allPages[i].features);
+        }
+        // return the modified first page
+        return {
+          featureCollection: { type: firstPage.type, features: firstPage.features },
+          on: {
+            click: (...args) => this.onAggregateDataClick(args)
+          },
+          total: firstPage.total
+        };
```
olzraiti, line 420: Collecting all the features with `allPages.reduce()` would be more clear (clear enough so that you wouldn't need the comments).

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -209,210 +237,272 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
-          )).pipe(tap(() => {
-            this.showingIndividualPoints = false;
-          }));
+          return this.getBoxDataOptions$(query).pipe(tap(() => this.showingIndividualPoints = false));
         }
-      }));
+      }),
+      // retry on timeout
+      timeout(WarehouseApi.longTimeout * 3),
+      delay(100),
```
olzraiti, line 488: What's the delay about? If it's necessary could you add some explanation into a comment for it.

olzraiti (changes requested):
I had missed the changes to `projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts` since Github hid it because there were so many changes. It was pretty clear, good job on cleaning the view port change detection. I left some minor wishes.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
Blodir: I actually don't know, I didn't change that line because I didn't know why it was there (git shows it as new line for some reason)

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -107,84 +172,47 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
+        draw: false
+      },
+      zoom: 1.5,
+      draw: false,
+      tileLayerName: LajiMapTileLayerName.openStreetMap,
+      clickBeforeZoomAndPan: true
+    };
     if ((environment as any).observationMapOptions) {
-      this._mapOptions = {...this._mapOptions, ...(environment as any).observationMapOptions};
+      Object.assign(this.mapOptions, ...(environment as any).observationMapOptions);
```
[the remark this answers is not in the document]
Blodir: Object reference is changed in the line above already, there's no reason to create another object (at small cost of performance) immediately after so mutation makes sense here.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
Blodir: Idk it might work, but I don't see why this would be an improvement? Calling detectChanges upon every view check is unnecessarily bad for performance, or alternatively there could be a flag to tell if it should be called on the next viewCheck, but that would complicate the code unnecessarily. An empty `setTimeout` isn't evil really, it just creates a callback in the js event loop. To be fair it's not at all obvious why it's required, but I think the alternatives are worse

olzraiti: It's not that bad as it is, we can go with this. I agree it would become more hard to follow with a flag. With the comment it's understandable.

In general I'm always worried about setTimeouts maintenance wise for many reasons. It's just a single tiny setTimeout which can seem innocent, but I don't like that it assumes that the table height is updated synchronously - if that changes this will break and it will be challenging to debug. It's difficult to find inline asynchronous operations in code. The `AfterViewChecked` should be guaranteened to run after the the actually is updated, even if it becomes async. Also, if there appear more setTimeouts, the order of execution can be really chaotic to follow.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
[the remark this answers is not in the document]
olzraiti: I think a reassignment is much more readable as the mutations done by any function are always pretty implicit and can be easily missed. The performance cost is really minimal...  But it's no big deal, I was just curious what was the reasoning.

Blodir:
The legend is in its own module now and the coordinate accuracy is added via content projection inside the observation-map component.

olzraiti (approved):
:+1: great!

PR #365, opened 2023-01-11
Improved obs map caching

Blodir:
* cache last 5 box queries. If viewport is contained within one of them, we don't need to do a new box query.
* finnish query used to just display 10km squares. Now it also displays single points when zoomed in


olzraiti (approved):
LGTM

PR #366, opened 2023-01-11
fix grc sample aggregate query failing

Blodir:
The regular warehouse queries no longer fail (however graphql queries do fail, that needs to be fixed in graphql-backend). The issue was that the query was targeting `/api/warehouse/query/sample/unit/aggregate` instead of `/api/warehouse/query/sample/aggregate`. I fixed it, but could not locate when/why this was changed (seeing as grc must have worked at some point)

https://www.pivotaltracker.com/story/show/184114619
https://184114619.dev.laji.fi/


olzraiti (approved):
LGTM, nice to see that tests were written.

PR #367, opened 2023-01-12
Fix notifications deletion bug

Blodir:
https://www.pivotaltracker.com/story/show/183164571

Note: could not reproduce the error locally, however this branch should fix the error that is described in Mikko's pictures. There's also some refactoring (it's 4 months old at this point I don't remember it too well I'm afraid)


olzraiti (commented):
The code seems to accept that notifications can be undefined. Why is it that the notifications can be undefined? I think it would simplify the code if we wouldn't accept that the notifications can be undefined.

olzraiti (approved):
LGTM

PR #369, opened 2023-01-19
Occurrence map interactivity

olzraiti:
* geometries are highlighted on hover
* active geometry is highlighted

### DIFF
--- a/projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
+++ b/projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
@@ -64,10 +64,12 @@ const getFeatureCollectionFromQueryCoordinates$ = (coordinates: any): Observable
   )
 );
 
+const classNamesAsArr = (c?: string) => c?.split(' ') || [];
+
 const getPointIcon = (po: PathOptions, feature: Feature): L.DivIcon => {
-  let currClassName = po.className;
+  let classNames = classNamesAsArr(po.className);
   const icon: any = L.divIcon({
-    className: currClassName,
+    className: classNames.join(' '),
     html: `<span>${feature.properties.count}</span>`
   });
   icon.setStyle = (iconDomElem: HTMLElement, po2: PathOptions) => {
@@ -75,13 +77,10 @@ const getPointIcon = (po: PathOptions, feature: Feature): L.DivIcon => {
     iconDomElem.style['height'] = '30px';
     iconDomElem.style['width'] = '30px';
     iconDomElem.style['border-radius'] = '100%';
-    if (po2.className) {
-      if (currClassName) {
-        currClassName.split(' ').forEach(c => iconDomElem.classList.remove(c));
-      }
-      po2.className.split(' ').forEach(c => iconDomElem.classList.add(c));
-      currClassName = po2.className;
-    }
+    const newClassNames = classNamesAsArr(po2.className);
+    classNames.forEach(c => iconDomElem.classList.remove(c));
+    newClassNames.forEach(c => iconDomElem.classList.add(c));
+    classNames = newClassNames;
   };
   return icon;
 };
--- a/projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
+++ b/projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
@@ -24,7 +24,8 @@ import { CollectionNamePipe } from '../../../shared/pipe/collection-name.pipe';
 import { LajiMapComponent } from '@laji-map/laji-map.component';
 import { LajiMapDataOptions, LajiMapOptions, LajiMapTileLayerName } from '@laji-map/laji-map.interface';
 import { PlatformService } from '../../../root/platform.service';
-import { DataOptions, DataWrappedLeafletEventData, TileLayersOptions } from 'laji-map';
+import { DataOptions, DataWrappedLeafletEventData, GetFeatureStyleOptions, TileLayersOptions } from 'laji-map';
+import { combineColors } from 'laji-map/lib/utils';
 import { environment } from '../../../../environments/environment';
 import { convertLajiEtlCoordinatesToGeometry, convertWgs84ToYkj, convertYkjToWgs, getFeatureFromGeometry } from '../../../root/coordinate-utils';
 import {
@@ -64,8 +65,9 @@ const getFeatureCollectionFromQueryCoordinates$ = (coordinates: any): Observable
 );
 
 const getPointIcon = (po: PathOptions, feature: Feature): L.DivIcon => {
+  let currClassName = po.className;
   const icon: any = L.divIcon({
-    className: po.className,
+    className: currClassName,
     html: `<span>${feature.properties.count}</span>`
   });
   icon.setStyle = (iconDomElem: HTMLElement, po2: PathOptions) => {
@@ -74,7 +76,11 @@ const getPointIcon = (po: PathOptions, feature: Feature): L.DivIcon => {
     iconDomElem.style['width'] = '30px';
     iconDomElem.style['border-radius'] = '100%';
     if (po2.className) {
-      iconDomElem.classList.add(po2.className);
+      if (currClassName) {
+        currClassName.split(' ').forEach(c => iconDomElem.classList.remove(c));
+      }
+      po2.className.split(' ').forEach(c => iconDomElem.classList.add(c));
+      currClassName = po2.className;
     }
   };
   return icon;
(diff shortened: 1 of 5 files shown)


projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -275,12 +306,14 @@ export class ObservationMapComponent implements OnChanges, OnDestroy {
...
       if (!dataOptions) { return; }
       const vis = this.visualization?.[this.visualizationMode];
-      if (vis?.getFeatureStyle) { dataOptions.getFeatureStyle = vis.getFeatureStyle; }
-      // changing mapData object reference so that laji-map recognizes that it has changed
-      this.mapData = [...this.mapData];
+      if (vis?.getFeatureStyle) {
+        dataOptions.getFeatureStyle = getFeatureStyleWithHoverAndActive(vis.getFeatureStyle);
+      }
+      // changing mapData object reference so that laji-map recognizes that it has changed.
+      this.mapData = this.lajiMap.map.getData();
```
olzraiti, line 316: The data at index `0` was the drawn box, index `1` hold the occurrence data. I'm not sure why this worked before, as the wrong `getFeatureStyle` was being updated.

I had to change the `mapData` reference to use the `getData()` from laji-map, or the data's `activeIdx` would been wrong (the getter return the data with the current value of the `activeIdx`, using the reference to laji-map data item doesn't since laji-map doesn't mutate the given data).

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -85,6 +91,30 @@ const BOX_QUERY_AGGREGATE_LEVELS = [
   ['gathering.conversions.wgs84Grid005.lat', 'gathering.conversions.wgs84Grid01.lon']
 ];
 
+
+const activeColor = '#6ca31d';
+
+const getFeatureStyleWithHoverAndActive = (getFeatureStyle: DataOptions['getFeatureStyle']) => (params: GetFeatureStyleOptions) => {
+  const {active, hovered} = params;
+  const style = getFeatureStyle(params);
+  const isActiveBox =params.feature.geometry.type !== 'Point' && active;
```
Blodir, line 100: missing space

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -85,6 +91,30 @@ const BOX_QUERY_AGGREGATE_LEVELS = [
...
+  const maxColorChangeDecimal = 100;
+  const color = combineColors(baseColor, ...(hovered ? ['#fff'] : []), maxColorChangeDecimal); // Highlight hovered item.
+  const className = [style.className, active && !isActiveBox ? 'laji-map-active-pointer' : undefined].filter(s => s).join(' ');
+  const _style = {...style, color, className};
+  if (isActiveBox) {
+    _style.weight = 3;
+    _style.color = color;
+    _style.fillColor = combineColors(style.color, activeColor, maxColorChangeDecimal); // Slide color towards active color.
+  }
+  return _style;
```
Blodir, line 115: I think currying is confusing in general. This could probably work better as a function decorator or something, but it's probably fine as it is.

Blodir: I'd probably call it augmentGetFeatureStyleWithHoverAndActive or decorateGetFeatureStyleWithHoverAndActive  or something like that since otherwise it's easy to confuse with the actual getFeatureStyle

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
Blodir: Good catch, the order of the data was changed in https://github.com/luomus/laji/commit/d3ebac5ac90549838a3fc385d3a15d961c531ce7

Blodir:
```
// changing mapData object reference so that laji-map recognizes that it has changed.
this.mapData = this.lajiMap.map.getData();
```
Now this line only implicitly changes the reference (it relies on `this.lajiMap.map.getData()` returning a new reference). It seems like it works so I guess that getData function is indeed returning a new reference (?), but the comment should be changed to reflect this

Blodir (changes requested):

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -74,7 +76,11 @@ const getPointIcon = (po: PathOptions, feature: Feature): L.DivIcon => {
     iconDomElem.style['width'] = '30px';
     iconDomElem.style['border-radius'] = '100%';
     if (po2.className) {
-      iconDomElem.classList.add(po2.className);
+      if (currClassName) {
+        currClassName.split(' ').forEach(c => iconDomElem.classList.remove(c));
+      }
+      po2.className.split(' ').forEach(c => iconDomElem.classList.add(c));
+      currClassName = po2.className;
```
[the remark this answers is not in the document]
olzraiti: The path options come from Leaflet - it'd be quite a lot of stabbing to make it work with a string array.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
olzraiti: I made it update the data directly to laji-map, instead of updating the local `mapData`, that is used only for passing new data to the template. I also tightened the occurrence map visualization typings, since it always has `getFeatureStyle`.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
olzraiti: A function decorator is just a curry function with syntactic sugar, and it couldn't be used here with the `@` syntax since it is done dynamically. I renamed it anyway, good idea.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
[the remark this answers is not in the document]
Blodir: I mean like this
```
const classNames = [];
// ...
icon = { ...
  className: classNames.join(' ')
// ...
if (classNames.length > 0) // line 79
  classNames.forEach => iconDomelem.classList.remove(...)
classNames = po2.className.split(' ')
classNames.forEach => iconDomElem.classList.add(...)
```
I think it's a bit cleaner that way

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
Blodir: Couldn't it? If say `getFeatureStyle` was a class method, then you annotate it with a method decorator `augmentHoverAndActive`. Then when passing getFeatureStyle to dataoptions you'd `.bind(this)`. I thought this would work, but I haven't tested it.

olzraiti: Well yes it could be done, but that would change the logic quite a bit as you then must have a single class method `getFeatureStyle`. I started trying that out, and in the end neither a curry fn or a decorator was needed. It escalated a lot... :D Here's a list of what changed:

* Only a single data options getter is used, both points & aggregate boxes use it.
* Points & aggregate box getters return just the feature collections. The box cache also caches just the feature collections, as the data options were anyway modified after a cache hit.  I think it's kind of more clean now that the data flows as feature collections, and the transformation into laji-map consumable data is done as late as possible.
* The feature click handler is done now by a proxy method that directs the event to the correct handler based on the feature type. Not necessarily an improvement, but a consequence of the other changes.

What do you think? I'm not completely sure if it seems better to me just because I wrote it and thus it's more comprehensible for me.

...The topic of whether to use a class method decorator vs a curry function  is a choice of personal preference to some extent. Both have use cases where the other might suite better, but in general I prefer simple currying for several reasons:
* more agnostic since it can be used for all functions, not just class methods
* for me they are much more easy to write - for a method decorator I'd have to look up the signature from the docs and even then it's just more complex (as a side note I think class decorators don't suffer from this though)
* it's plain js, where as decorators are an experimental feature. I've actually ran into problems with them when I've changed the TS compiler on laji-map (I think the execution order of the method decorators was different between compilers). If used only inside an Angular project, this isn't a problem of course.

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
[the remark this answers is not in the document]
olzraiti: Right, yeah that makes sense

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
Blodir: Looks good! I think this is definitely the right approach.

About curry vs decorator: The thing I like about decorators is that the behavior is explicit. When I see a decorator I know immediately that it takes a function, returns a function etc. All of your reasons are totally fair and true though.

Blodir (approved):
LGTM!

PR #370, opened 2023-01-25
Occurrence map changing projection shows only small portion of data

Blodir:
Fixes: https://www.pivotaltracker.com/story/show/184291899

Also all occurrences of laji-map "size invalidation" were removed since it was deemed redundant.


olzraiti (approved):
LGTM

PR #371, opened 2023-01-31
Fix label designer pdf 184051702

aorin:
https://www.pivotaltracker.com/story/show/184051702

This pull request fixes label designer two-sided printing. The printing didn't work properly with the css property `direction: rtl` so I used another method to achieve the same result.


projects/label-designer/src/lib/pipe/label-backside-sort-pipe.ts
```diff
@@ -0,0 +1,23 @@
...
+    while (sortedData.length < dim.cols * dim.rows) {
+      stopIdx = startIdx + dim.rows - 1;
+      for (let i = startIdx; i <= stopIdx; i++) {
+        sortedData.push(data[i]);
+      }
+      startIdx -= dim.rows;
+    }
+
+    return sortedData;
+  }
```
olzraiti, line 22: A while loop with an inner for loop, manipulating two indices all at the same time, yikes :) It's difficult to understand what's going on here.

1. It would help if there'd be some description how the sorting should work ("back side sorting" implies that it's the order in the back of the paper, but that doesn't tell how they should be ordered in the back side).

2. I think doing implementing this with "functional" operators like `reduce`, `map` etc could help, if possible. Usually you can eliminate (a) the usage of manually tracked indices that way and (b) while loops which can cause major bugs since they freeze the whole process if they never finish.

olzraiti (changes requested):
The sorting function feels a bit dangerous, but otherwise the PR looks good.

olzraiti:
BTW there's a merge conflict

projects/label-designer/src/lib/pipe/label-backside-sort-pipe.ts
aorin: I tried to simplify it, now it only has one while loop. I also added the description

olzraiti: Thanks! The description helps a lot, and the iteration is more readable.

olzraiti (approved):
:+1:

PR #372, opened 2023-01-31
fix obs search filters date change management [fixes #184323833]

olzraiti:
https://www.pivotaltracker.com/story/show/184323833
https://184323833.dev.laji.fi/observation/statistics?countryId=ML.206&qualityIssues=BOTH&firstLoadedSameOrAfter=2023-01-20&observerPersonToken=true

The laji-datepicker was used with two-way binded ng model (`[(banana in a box syntax)]`), and it also had a change event listener. That didn't make sense, so it made it use only the two-way binded ng model. It had broken after the datepicker refactoring which made the datepicker implement ngModel value accessor interface better.

### DIFF
--- a/projects/laji/src/app/shared/datepicker/datepicker.component.ts
+++ b/projects/laji/src/app/shared/datepicker/datepicker.component.ts
@@ -137,6 +137,7 @@ export class DatePickerComponent implements ControlValueAccessor {
     this.calendarUIValue = value;
     this.dateSelect.next(this.value);
     if (prevValue !== this.value) {
+      console.log('change', this.value, !!this.onChangeCallback);
       this.onChangeCallback?.(this.value);
       this.onTouchedCallback?.();
     }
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/shared/datepicker/datepicker.component.ts
```diff
@@ -137,6 +137,7 @@ export class DatePickerComponent implements ControlValueAccessor {
     this.calendarUIValue = value;
     this.dateSelect.next(this.value);
     if (prevValue !== this.value) {
+      console.log('change', this.value, !!this.onChangeCallback);
```
aorin, line 140: This log could be removed?

olzraiti: oopsie. Removed

aorin (approved):
Looks good

PR #373, opened 2023-01-31
Observation map data opacity [#184335037]

olzraiti:
https://www.pivotaltracker.com/story/show/184335037
https://184335037.dev.laji.fi/observation/map

* Adds observation data layer opacity controller to the observation map.
* Removes observation map legend opacity

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13211,9 +13211,9 @@
       }
     },
     "laji-map": {
-      "version": "3.17.11",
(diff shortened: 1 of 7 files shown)


rpulkka (approved):
Approving. The code adds a setting to the map which allows the user to change the observation visualization overlay's opacity. Opacity has been converted to hex code. I do not see any issues with the code.

PR #376, opened 2023-02-06
Fix front page obs map [#184370599]

olzraiti:
https://www.pivotaltracker.com/story/show/184370599
https://184370599.dev.laji.fi

The obs map on the front page redirects to /observation/map on click. The map view is also locked (can't be panned/zoomed).

### DIFF
--- a/projects/laji/src/app/+home/home.component.html
+++ b/projects/laji/src/app/+home/home.component.html
@@ -72,19 +72,20 @@
       </a>
     </div>
     <div class="map-container">
-      <laji-observation-map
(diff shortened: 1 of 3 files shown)


aorin (commented):
The code looks ok but the map could look more like a link. The cursor could be a pointer or maybe the whole map could be wrapper inside a link element like the image below the text " Help identify observations". I'm also wondering if the link should have the same filters as the link below the map

olzraiti:
Good idea, I wrapped the map inside a link element and made it use the same filter as the link below the map.

aorin (approved):
Looks good

PR #377, opened 2023-02-07
Fix biology tab descriptions & Refactor CheckLangService [#184388410]

rpulkka:
https://184388410.dev.laji.fi/taxon/MX.205601/biology
https://www.pivotaltracker.com/n/projects/2346653/stories/184388410

In some cases, some descriptions were not rendered on taxon card's biology tab. Turned out that the reason was a condition that prevented showing the whole description group if any number of its variables were not translated into the used language (line 32 of taxon-biology-component.html). Now the condition for showing the group passes, if at least one variable has translations in the used language. Each variable has their own check now, and is shown if there's content in the used language.

On top that, this PR includes a refactor of CheckLangService, which seemed to have a lot of unused variables and functionalities, so I deleted those and renamed variables and functions to be more readable. CheckLangService is used to create a translationChecklist, which is used to check which variables have translations in the used language, while rendering variables in the template.

### DIFF
--- a/projects/laji/src/app/+taxonomy/taxon/info-card/taxon-biology/taxon-biology.component.html
+++ b/projects/laji/src/app/+taxonomy/taxon/info-card/taxon-biology/taxon-biology.component.html
@@ -29,7 +29,7 @@
         <div class="panel-group" role="tablist" aria-multiselectable="true">
           <div *ngFor="let group of taxonDescription[activeDescription]?.groups; let i = index">
             <div class="panel panel-default"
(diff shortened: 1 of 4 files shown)


MLumme (commented):
Otherwise this seems fine, but in check-lang.service.ts couldn't you put perhaps a boolean variable in the group at line 40 (new version), which would act like the old totalVisible, but for a group,  to signify that the group has at least one visible text, and use that in taxon.biology.component. html line 32, instead of filtering the group values there?

rpulkka:
Sounds like an improvement, fixed now.

MLumme (approved):
Look fine, approving

PR #378, opened 2023-02-07
Polygon occurrence search [#181997072]

olzraiti:
Polygon search was already implemented once, but there were performance problems to we reverted it. This PR reverts the reverted PR. So the code has been reviewed already, but there was quite a big merge conflict so it might be useful to look through this anyway.

Also, this isn't a mere merge only, but the coordinates intersection logic was removed also since the backend doesn't support that anymore.

The revert flow is like this:
* #283 was the original PR
* #295 reverted most of #283, keeping some refactoring
* This reverts #295 and removes the coordinates intersection logic & refactors coordinate utils

There were some typing problems, which escalated in a bit of changes in the `coordinate-utils`. There are some changes that could break something, but if they do, they are probably currently working in some very buggy, unstable way.

For example:
* `getFeatureFromGeometry` could return a feature with an empty object as the geometry. That's not valid GeoJSON, and if we are doing that somewhere, we really shouldn't be doing so. There's some unsafe error management going on if this breaks something.
* `convertLajiEtlCoordinatesToGeometry` could return `undefined` for a falsy input string. The function is recursive, so you could end up with a `GeometryCollection` with undefined geometries, which isn't valid JSON. If there's code relying on this function returning undefined, that's also unsafe error management. I made the functino throw an error if the input doesn't procude any sane geometry. Also, the function returned the geometry coordinates as strings, which isn't valid GeoJSON and since the function was missing typings, it could result in some bugs if the code elsewhere assumes the coordinates to be numbers.


projects/laji/src/app/+observation/result/observation-result.component.ts
```diff
@@ -143,35 +151,63 @@ export class ObservationResultComponent implements OnInit, OnChanges {
...
+        }
       } else {
         query.coordinates = undefined;
+        query.polygonId = undefined;
       }
     });
     this.queryChange.emit(query);
   }
 
+  registerPolygon$(polygon: any) {
```
Blodir, line 194: Is this G.Polygon? Could it be typed?

projects/laji/src/app/shared/api/WarehouseApi.ts
```diff
@@ -483,6 +483,15 @@ export class WarehouseApi {
     return this.http.get(path, {params: queryParameters});
   }
 
+  public registerPolygon(polygon: any, personToken: string, crs: string) {
```
Blodir, line 486: Again, is this G.Polygon? Could it be typed?

projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.ts
```diff
@@ -49,19 +50,12 @@ interface AggregateQueryResponse {
...
 
 // Given coordinates in warehouse query format
-// Returns a featureCollection visualizing that set of coordinates
-const getFeatureCollectionFromQueryCoordinates$ = (coordinates: any): Observable<any> => (
+// Returns features visualizing that set of coordinates
+const getFeaturesFromQueryCoordinates$ = (coordinates: string[]): Observable<G.Feature<G.Polygon>[]> => (
   ObservableOf(coordinates
     ? coordinates.map(
-      (coord: any) => getFeatureFromGeometry(convertLajiEtlCoordinatesToGeometry(coord))
+      (coord) => getFeatureFromGeometry(convertLajiEtlCoordinatesToGeometry(coord))
```
Blodir, line 57: If we had `strict` typing enabled, tsc would complain about implicit any, so we should probably keep the type in there.

Blodir (approved):
Looks ok, though I didn't find it super easy to follow. There's some missing types, but there are already missing types so the PR doesn't really make things any worse in that regard.

olzraiti:
Those any's were there probably because I didn't dare to touch the coordinate utils in the previous iteration if this feature. I added typings for those now

PR #380, opened 2023-02-15
Fix species download getting stuck

aorin:
https://www.pivotaltracker.com/story/show/184256724
https://184256724.dev.laji.fi/en/taxon/list?onlyFinnish=true

Species list download used to get stuck easily because it executed a function called `dataToLookup` way too many times. I moved the function to `base-data.service` and added caching.


MLumme (approved):
Looks good, approved.

PR #381, opened 2023-02-15
Fix data bank delete

aorin:
https://www.pivotaltracker.com/story/show/184274259
https://184274259.dev.laji.fi/theme/datasets

This fixes data importer complaining about required columns when user tries to delete data. Now it doesn't check the required columns if there is a "delete" column in the file.


MLumme (approved):
Looks ok, approved.

PR #383, opened 2023-02-15
LajiFormBuilder

olzraiti:
https://210820.dev.laji.fi/project-edit

This PR adds the new `/project-edit` module, where Vihko forms can be edited.

`LajiFormBuilder` is a React app, that is rendered on the page similarly as `LajiForm`. It allows editing forms via a GUI, that is displayed in the bottom as a fixed window on top of the project page.

There isn't a Pivotal task, as the development hasn't been done against some requirements, but we had a meeting about the app with Mikko and Eppu and they approved that it should be brought to production.

Some notes about the code:
* `project-form-option` directive is used so that the "magnet" tool can target the elements and show the corresponding form options
* `project-form` holds a local instance of the form, which the `project-form-edit` updates when the form is edited
* `project-form-edit` has access to the `project-form` so that it can rerender the `LajiForm` component (when the refresh button is clicked). The tool is there because you can easily break the  UI but you can recover it then.


projects/laji/src/app/shared/service/project-form.service.ts
```diff
@@ -31,16 +31,48 @@ export interface NamedPlacesRouteData extends NamedPlacesQueryModel {
...
+   * LajiFormBuilder can change the language of the form, without changing the lang of the whole page.
+   */
+  public localLang$ = new BehaviorSubject<string>(this.translate.currentLang);
+  public remountLajiForm$ = new Subject<void>();
+
+  getFormFromRoute$(route: ActivatedRoute): Observable<Form.SchemaForm> {
+    return this.getFormID(route).pipe(switchMap(formID => this.getForm(formID)));
+  }
+
+  getForm(id: string): Observable<Form.SchemaForm> {
```
aorin, line 55: The naming could be more consistent, some have `$` in their name (`getFormFromRoute$`) and some don't (`getForm`)

projects/laji/src/app/shared/directive/project-form-option.directive.ts
```diff
@@ -0,0 +1,44 @@
...
+    });
+    el.className.trim();
+  }
+
+  get lajiFormOption() {
+    return this._formOption ? this._formOption.join(' ') : '';
+  }
+}
+
+function singleFormOptionToClassName(o: string) {
```
aorin, line 38: Could have return type

projects/laji/src/app/+project-form/about/about.component.html
```diff
@@ -1,25 +1,28 @@
 <laji-technical-news></laji-technical-news>
 
-<ng-container *ngIf="aboutData$ | async as data">
+<div *ngIf="aboutData$ | async as data" lajiFormOption="mobile dataset about collectionID options.disableRequestDescription">
```
aorin, line 3: It looks to me that `mobile dataset about` should be `options.mobile options.dataset options.about` since they are under the options object

aorin: Because these are just strings and the types are not used, it's easy to make mistakes. It would be better if it would use types but I don't know if it would take too much work at this point. It could look something like this:

This file:
```
<div *ngIf="aboutData$ | async as data" [lajiFormOption]="['collectionID']" [lajiFormOptionOption]="['mobile', 'dataset' , 'disableRequestDescription']">
```

Directive:
```
selector: '[lajiFormOption], [lajiFormOptionOption]'
...
@Input() lajiFormOption: (keyof SchemaForm)[];
@Input() lajiFormOptionOption: (keyof FormOptions)[];
ngOnChanges() { // do something with the options }
```

aorin (changes requested):

projects/laji/src/app/+project-form/about/about.component.html
olzraiti: It's a nice idea, but with this approach we'd have to have a different directive for each depth level, which would be messy. For example named place options are in an object inside the form options. Currently those set simply with e.g. `options.namedPlaceOptions.useLabel`.

It's easy to make mistakes yes consequences of a typo aren't that dangerous. The syntax is also less bloaty in the current way it's done.

olzraiti:
Thanks for the review! I addressed the issues.

aorin (approved):
Thanks, looks good

PR #384, opened 2023-02-16
Bookmarking bug fix #2

Blodir:
https://www.pivotaltracker.com/story/show/182564517

Previously when I fixed this issue I had forgotten line 604 somehow so it was reverted. Should work as expected now.


rpulkka (approved):
Approving. PR contains missing assignment to "this.query" after modifying it. The bug is fixed in the feature branch.

PR #387, opened 2023-02-16
Alive filter 183862013

aorin:
https://183862013.dev.laji.fi/observation/list?alive=true
https://www.pivotaltracker.com/story/show/183862013

Adds a new filter to observation search


projects/laji/src/i18n/fi.json
```diff
@@ -2079,4 +2082,4 @@
   "whatsNew.message": "laji.fi on Suomen Lajitietokeskuksen testikäytössä oleva portaali. Tietoja täydennetään ja ominaisuuksia parannetaan jatkuvasti. Palaute on tervetullutta - sitä voi lähettää palautelomakkeella tai osoitteeseen info@laji.fi. Kiitos!<p>Voit katsoa tuoreimmat muutokset <a href=\"https://laji.fi\">etusivun</a> ajankohtaista osiosta.</p",
   "wholeYear": "Koko vuosi {{year}}",
   "yes": "Kyllä"
-}
```
rpulkka, line 2082: Do we have guidelines on whether to use newline at the end of files or not?

rpulkka (approved):
Approving. No issues in the code, and feature works as expected.

projects/laji/src/i18n/fi.json
aorin: We don't have any general guidelines but I think most of the time we have newline at the end

PR #388, opened 2023-02-16
Add button for minimizing the occ viz legend

Blodir:
https://www.pivotaltracker.com/story/show/184369502


rpulkka (approved):
Code looks good to me, approving. The feature works like intended.

PR #389, opened 2023-02-16
Fix collection filter 184003345

aorin:
https://184003345.dev.laji.fi/observation/statistics
https://www.pivotaltracker.com/story/show/184003345

The reason for this bug was that the `collections$` observable (in`projects/laji/src/app/shared-modules/collections-select/collections-select.component.ts`) was recreated when the query changed and the tree select component subscribed only the first observable. It's preferred that the observables are created only once so I changed that for both collection select and informal group select. I also added some caching.


aorin:
I ended up refactoring quite a much so I think you @MLumme are the best person to check that this didn't break anything

MLumme (approved):
Refactoring looks to be a great improvement and the original issue is fixed, approved.

PR #393, opened 2023-03-15
Rm doc table from doc form [#184379680]

olzraiti:
https://184379680.dev.laji.fi/project/MHL.45
https://www.pivotaltracker.com/story/show/184379680

Removes the document table from document form.

The difference to dev can be seen by navigating to some lolife document that has prior observations.


aorin (approved):
:+1:

PR #395, opened 2023-03-15
Fix modal height

aorin:
https://183943758.dev.laji.fi/observation/list
https://www.pivotaltracker.com/story/show/183943758

Fixes problems with modals, especially with informal group and source filter modals where the "ok" button is hard to see. This PR makes the modals to adapt to screen height and´the modal body scrollable.


rpulkka (approved):
Good work, approving. The code adds the .flex-container style for the tree-select-modal-component, and the general styles that make the modals scrollable.

PR #400, opened 2023-03-29
BSG fine-tuning

aorin:
Can be tested in: 
https://kerttu-global-dev.laji.fi/identification/results

Small changes to the bsg:
- added filters to the species result list
- added Crowdin scripts (BSG has a separate Crowdin project)


projects/kerttu-global/src/app/+identification/identification-results/identification-results.component.ts
```diff
@@ -22,16 +23,23 @@ export class IdentificationResultsComponent implements OnInit {
   siteStats$: Observable<IIdentificationSiteStat[]>;
   userStats$: Observable<IIdentificationUserStatResult>;
   speciesStats$: Observable<IIdentificationSpeciesStat[]>;
+  ownSpeciesStats$: Observable<IIdentificationSpeciesStat[]>;
```
Blodir, line 26: Not a problem with this PR specifically, but I don't like using hungarian notation for interfaces. Also Angular, or any other lib we use doesn't do it either. Eg. this one should just be `IdentificationSpeciesStat` imo. Maybe this issue should be raised during dev talk sometime.

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -389,12 +389,6 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnChanges, OnD
   }
 
   private defaultSort(rows: any[]): any[] {
```
Blodir, line 391: Would be trivial to add generics here
```
  private defaultSort<T>(rows: T[]): T[] {
    return [...rows].sort((a: T, b: T) => a.preSortIndex < b.preSortIndex ? -1 : 1);
  }
```

Blodir (changes requested):
Adding generics would be nice, otherwise looks good imo

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
aorin: It can't be a generic type because of the `preSortIndex` prop but I added a new type `DatatableRow`

projects/kerttu-global/src/app/+identification/identification-results/identification-results.component.ts
aorin: Sounds good to me, it would be better if the same notation would be used everywhere

Blodir (approved):

PR #401, opened 2023-03-30
Np list/info modal use schema [#184683726]

olzraiti:
https://184683726.dev.laji.fi/project/MHL.45/form/MHL.45/places
https://www.pivotaltracker.com/story/show/184683726

Named places list and the info modal uses labels/decision to use a label pipe from the named place form & document form schemas.

### DIFF
--- a/projects/laji/src/app/+project-form/form/named-place/np-choose/np-list/np-list.component.ts
+++ b/projects/laji/src/app/+project-form/form/named-place/np-choose/np-list/np-list.component.ts
@@ -21,6 +21,7 @@ import { ObservationTableColumn } from '../../../../../shared-modules/observatio
 import { DatatableColumn } from '../../../../../shared-modules/datatable/model/datatable-column';
 import { DatatableComponent } from '../../../../../shared-modules/datatable/datatable/datatable.component';
 import { SelectionType, SortType } from '@swimlane/ngx-datatable';
+import { NpInfoComponent } from '../../np-info/np-info.component';
 
 @Component({
   selector: 'laji-np-list',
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/+project-form/form/named-place/np-info/np-info.component.ts
```diff
@@ -99,7 +106,7 @@ export class NpInfoComponent implements OnInit, OnChanges, AfterViewInit {
       }
 
       let pipe;
-      if (field === 'taxonIDs') {
+      if (field === 'taxonIDs' || (fieldSchema?.type === 'string' && fieldSchema.oneOf)) {
```
aorin, line 109: This conditition `fieldSchema?.type === 'string' && fieldSchema.oneOf` was also used before. It could be moved to some service as a function `isStringEnumField(schema): boolean`

projects/laji/src/app/+project-form/form/named-place/np-choose/np-list/np-list.component.ts
```diff
@@ -165,27 +172,42 @@ export class NpListComponent implements OnDestroy {
...
     }
-    const cols: ObservationTableColumn[] = [];
-    for (const path of this._fields) {
-      const {cellTemplate, ...columnMetadata} = this.columnsMetaData[path] || {} as DatatableColumn;
-      const col: DatatableColumn = {
+
+
+    const cols = this._fields.reduce((_cols, path) => {
+
+      const splits = path.split('.');
```
aorin, line 179: It looks to me that this does about the same than the lines 87-97 in the `np-info.component` but maybe there are differences? I think it would be clearer if all the schema parsing stuff was moved to some separate service since they make the functions long and little hard to read. This would become `const schema = someService.getSchemaFromPath(path)`

aorin (changes requested):
Looks good otherwise but I feel the functions could be broken into a little smaller pieces

projects/laji/src/app/+project-form/form/named-place/np-info/np-info.component.ts
olzraiti: There are only a few checks for the string enum schema type in the repo, and I didn't come up with a good place where to put the function. Creating a whole new schema utils for this simple check seems excessive. I binded the check into variables called `isEnumSchema`, should help readibility a bit.

projects/laji/src/app/+project-form/form/named-place/np-choose/np-list/np-list.component.ts
olzraiti: Good idea, refactored the schema parsing into a single static function.

projects/laji/src/app/+project-form/form/named-place/np-info/np-info.component.ts
aorin: Ok, it sounded like a common check but maybe it's ok like that for now

aorin (approved):
Thanks for the changes!

PR #402, opened 2023-04-18
Observation search changes

aorin:
https://www.pivotaltracker.com/story/show/184017766

Changes to the observation search:
 - Adds a search button to the sidebar which updates the filters
 - Shows an instruction page when no tab is selected and no tab is selected as a default


projects/laji/src/app/+observation/form/observation-form.component.scss
```diff
@@ -1,10 +1,6 @@
 @import '../../../styles/variables';
 @import 'projects/laji-ui/src/lib/vars';
 
-:host ::ng-deep laji-observation-active .open-select {
-  border-bottom: 3px solid rgba(0,0,0,0)
```
olzraiti, line 5: semicolon missing

projects/laji/src/app/+observation/observation.facade.ts
```diff
@@ -100,7 +95,7 @@ export class ObservationFacade {
     private warehouseApi: WarehouseApi,
     private observationDataService: ObservationDataService
   ) {
-    this.updateState({..._state, ...this.persistentState});
+    this.updateState({..._state});
```
olzraiti, line 98: This is a bit out of the scope if this PR, but do you know why we have the singleton `_state` variable? The variable outlives the component lifecycle, so if you e.g. navigate away from the obs page and navigate back, the component state is initialized here with the `_state` var which could still hold the old state. I don't know why we haven't ran into bugs with this, are we maybe clearing the `_state` when the component page is destroyed? Regardless of how it manages to work, I don't see any reason why it's a singleton. If there's no reason for having it, I anticipate the logic would be simpler without it.

projects/laji/src/app/+observation/observation.facade.ts
```diff
@@ -126,17 +121,26 @@ export class ObservationFacade {
...
-        this.hashCache['query'] = hash;
+        this.queryHash = hash;
 
-        this.updateState({..._state, query, loadingUnits: true, loadingTaxa: true});
+        this.updateState({..._state, query, newQuery, newQueryHasChanges: false, loadingUnits: true, loadingTaxa: true});
       })
     );
   }
 
+  updateNewQuery(warehouseQuery: WarehouseQueryInterface) {
```
olzraiti, line 139: The naming of these could also be more explicit, took me a while to understand the difference between these:

`updateQuery` -> `updateActiveQuery`
`updateNewQuery` -> `updateTmpQuery`

projects/laji/src/app/+observation/observation.facade.ts
```diff
@@ -126,17 +121,26 @@ export class ObservationFacade {
...
-        if (this.hashCache['query'] === hash) {
+        if (this.queryHash === hash) {
+          this.updateState({..._state, newQuery, newQueryHasChanges: false});
           return;
         }
-        this.hashCache['query'] = hash;
+        this.queryHash = hash;
 
-        this.updateState({..._state, query, loadingUnits: true, loadingTaxa: true});
+        this.updateState({..._state, query, newQuery, newQueryHasChanges: false, loadingUnits: true, loadingTaxa: true});
```
olzraiti, line 134: The state change between the if-condition and the "else part" would be easier to understand if a single state object would be binded to a variable before the if-block, and the "else" part would add stuff to the state var.
```
const nextState ={..._state, newQuery, newQueryHasChanges: false};
       if (this.queryHash === hash) {
          this.updateState(nextState);
          return;
        }
        this.queryHash = hash;

        this.updateState({...nextState, loadingUnits: true, loadingTaxa: true});
```

projects/laji/src/app/+observation/result/observation-result.component.ts
```diff
@@ -61,6 +61,7 @@ export class ObservationResultComponent implements OnInit, OnChanges {
   @Input() resultBase: 'unit'|'sample' = 'unit';
   @Input() basePath = '/observation';
   @Input() query: WarehouseQueryInterface = {};
+  @Input() newQuery: WarehouseQueryInterface = {};
```
olzraiti, line 64: Naming scheme like described in prev comments

projects/laji/src/app/+observation/result/observation-result.component.ts
```diff
@@ -70,6 +71,7 @@ export class ObservationResultComponent implements OnInit, OnChanges {
   @Input() listSettings: ISettingResultList;
 
   @Output() queryChange = new EventEmitter<WarehouseQueryInterface>();
+  @Output() newQueryChange = new EventEmitter<WarehouseQueryInterface>();
```
olzraiti, line 74: Naming scheme like described in prev comments

projects/laji/src/app/+observation/result/observation-result.component.ts
```diff
@@ -150,46 +150,63 @@ export class ObservationResultComponent implements OnInit, OnChanges {
...
+        if (e.features.length === 1) {
+          const newCoordinates = this.newQuery.coordinates || [];
+          const oldCoordinates = this.query.coordinates || [];
+          if (!Util.equalsArray(newCoordinates, oldCoordinates) || this.newQuery.polygonId !== this.query.polygonId) {
+            this.newQueryChange.emit({ ...this.newQuery, coordinates: this.query.coordinates, polygonId: this.query.polygonId });
+          }
+        } else if (e.features.length > 1) {
+          throw new Error('Something wrong with map, there should never be multiple deletable geometries');
+        }
+        return;
```
olzraiti, line 177: Could help readibility, if the `e.features.length > 1` check would be on top, so there'd be no need for the other if block. The return statement would be more clear also since it would be inside the same block.

projects/laji/src/app/+observation/search-query.service.ts
```diff
@@ -344,4 +345,37 @@ export class SearchQueryService implements SearchQueryInterface {
   public queryUpdate(data = {}): void {
     this.queryUpdatedSource.next(data);
   }
+
+  public static hasValue(value): boolean {
+    const type = typeof value;
+    return (type !== 'undefined' && (type === 'boolean' || type === 'number' || (value && value.length > 0)));
+  }
```
olzraiti, line 352: Add typing for the `value` input, otherwise if this is used for an object, it will return false (I assume it never receives an object in it's use case)

projects/laji/src/app/+observation/search-query.service.ts
```diff
@@ -344,4 +345,37 @@ export class SearchQueryService implements SearchQueryInterface {
...
+
+  public static getDifferenceBetweenQueries(query1: WarehouseQueryInterface, query2: WarehouseQueryInterface): WarehouseQueryInterface {
+    const query1Keys = Object.keys(query1);
+    const query2Keys = Object.keys(query2);
+    const query2UniqueKeys = query2Keys.filter(k => !query1Keys.includes(k));
+    const uniqueKeys = query1Keys.concat(query2UniqueKeys);
```
olzraiti, line 358: Could use a `Set`: `const uniqueKeys = new Set(query1Keys.concat(query2Keys))`

projects/laji/src/app/+observation/search-query.service.ts
```diff
@@ -344,4 +345,37 @@ export class SearchQueryService implements SearchQueryInterface {
...
+        return;
+      }
+
+      const areArrays = Array.isArray(value1) && Array.isArray(value2);
+      if (!(value1 === value2 || (areArrays && Util.equalsArray(value1, value2)))) {
+        changed[key] = query2[key];
+      }
+    });
+    return changed;
+  }
```
olzraiti, line 375: `reduce` instead of `forEach` would help understanding what is the purpose of the iteration (=collecting the changed keys)

projects/laji/src/app/+observation/view/observation-view.component.ts
```diff
@@ -151,17 +147,16 @@ export class ObservationViewComponent implements OnInit, OnDestroy {
...
   }
 
-  empty() {
-    this.observationFacade.clearQuery();
-    this.form.empty();
-  }
-
-  toggleInfo() {
-    this.observationFacade.toggleIntro();
+  updateNewQuery(query: WarehouseQueryInterface, showSidebarOnMobile = false) {
```
olzraiti, line 150: Naming scheme like described in prev comments

projects/laji/src/app/+observation/view/observation-view.component.ts
```diff
@@ -151,17 +147,16 @@ export class ObservationViewComponent implements OnInit, OnDestroy {
...
+    if (showSidebarOnMobile) {
+      this.sidebar.showOnMobile();
+    }
+    this.observationFacade.updateNewQuery({...query});
   }
 
-  onQueryChange(event: WarehouseQueryInterface) {
-    this.observationFacade.updateQuery$(event).subscribe();
+  updateQuery(query: WarehouseQueryInterface) {
```
olzraiti, line 157: Naming scheme like described in prev comments

olzraiti (changes requested):
It's a bit hard to follow due to the weight of history of the code. The singleton state that the facade uses seems really suspicious and abuses the facade pattern.

I left some pretty minor change requests. Good job!

projects/laji/src/app/+observation/observation.facade.ts
aorin: Hmm, can't think any reason for it. Might be an old relic. I guess it doesn't cause any problems since it sets the correct state when it loads the query from url parameters

aorin:
Thanks for good suggestions! I think I have done all the changes now. I also found the facade and some other parts quite confusing, it would need refactoring at some point

projects/laji/src/app/+observation/form/observation-form.component.scss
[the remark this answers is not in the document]
olzraiti: still missing :)

olzraiti (approved):
Thanks!

projects/laji/src/app/+observation/form/observation-form.component.scss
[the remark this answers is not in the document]
olzraiti: Approved anyway, doesn't matter.

PR #403, opened 2023-04-18
Added a marker to denote primary specimen to document viewer.

MLumme:
The marker should always stay next to the first row of the unit. Testable at https://183999568.dev.laji.fi/view?uri=http:%2F%2Fid.luomus.fi%2FJA.1122334456

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM node:14
 
 # set working directory & user
-USER node
+# USER node
 WORKDIR /app
 
 # prepare the env (npm will not run script if using root)
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@
 FROM node:14
 
 # set working directory & user
-# USER node
+USER node
 WORKDIR /app
 
 # prepare the env (npm will not run script if using root)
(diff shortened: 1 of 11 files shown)


Dockerfile
```diff
@@ -2,7 +2,7 @@
 FROM node:14
 
 # set working directory & user
-USER node
+# USER node
```
olzraiti, line 5: Is this committed on purpose?

projects/laji/src/app/shared-modules/document-viewer/unit-rows/unit-rows.component.html
```diff
@@ -1,15 +1,34 @@
+<ng-template #primarySpecimenHideFacts>
+  <ng-container *ngIf="!showFacts then primarySpecimen"></ng-container>
+</ng-template>
+<ng-template #primarySpecimen>
+  <ng-container *ngIf="unit.primarySpecimen">
+    <div class="primary-specimen">{{"result.unit.primarySpecimen" | translate}}</div>
```
olzraiti, line 6: use white space in `{{ }}`

olzraiti (changes requested):

Dockerfile
MLumme: It definitely is not supposed to be there, it's just to get the thing to run locally in docker, but I must've been careless when adding the files.

MLumme: Both should be fixed now in the latest commit.

olzraiti (approved):
Thanks!

PR #404, opened 2023-05-04
Observation search alert

aorin:

The page shows an info toast about the new search button after the user touches the observation filters.

This also fixes the style of the sidebar in the species card (https://dev.laji.fi/taxon/MX.46587?showTree=true) that was broken by the observation search pull request.


olzraiti (approved):
LGTM :+1:

PR #406, opened 2023-05-10
Observation search alert fine-tuning

aorin:

Changes the observation search info toast so that it is shown only once. I did that by bringing back the `persistentState` to `ObservationFacade`. This pull request also removes the coordinate info toast that was not needed anymore.


olzraiti (approved):
LGTM

PR #409, opened 2023-05-23
CollectionAndRecordquality filter values not properly set 184654630

MLumme:
Story: https://www.pivotaltracker.com/story/show/184654630, testable here https://184654630.dev.laji.fi/observation/list?collectionAndRecordQuality=PROFESSIONAL:EXPERT_VERIFIED,COMMUNITY_VERIFIED,NEUTRAL,UNCERTAIN;HOBBYIST:EXPERT_VERIFIED,COMMUNITY_VERIFIED,NEUTRAL;AMATEUR:EXPERT_VERIFIED,COMMUNITY_VERIFIED, problem was that for laji.fi to set the filters properly that last COMMUNITY_VERIFIED needed to be followed by a semicolon, now it works with or without it.

### DIFF
--- a/projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
+++ b/projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
@@ -238,25 +238,25 @@ export class SelectSubcategoriesComponent implements OnChanges {
           const splitParamGlobal = param.split(',');
           this.loopInsideOptions(this.subCategories, this.options, splitParamGlobal, false);
         } else {
-          const splitParamCategories = this.arrFromUrlString(param);
+          const splitParamCategories = arrFromUrlString(param);
           this.loopInsideOptions(this.subCategories, this.options, splitParamCategories, true);
         }
       } else {
         this.unselectedOptions = this.unselectedOptions ? this.unselectedOptions : {};
       }
     }
-  }
 
-  private arrFromUrlString(urlString: string): string[] {
-    const rebuilt = urlString.split(';').filter(s => s.length > 0);
-    const finalRebuilt = [];
+    function arrFromUrlString(urlString: string): string[] {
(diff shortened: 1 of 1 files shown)


projects/laji/src/app/shared-modules/select/select-subcategories/select-subcategories.component.ts
```diff
@@ -248,7 +248,7 @@ export class SelectSubcategoriesComponent implements OnChanges {
   }
 
   private rebuiltParamSubCategory(urlString) {
-    const rebuilt = urlString.slice(0, -1).split(';');
+    const rebuilt = urlString.charAt(urlString.length - 1) === ';' ? urlString.slice(0, -1).split(';') : urlString.split(';');
```
Blodir, line 251: This would be a bit cleaner:
```
const rebuilt = urlString.split(';').filter(s => s.length > 0);
```

Blodir: Also the function could be renamed to something like `arrFromUrlParams` and have corresponding typing.

Blodir: Also, imo utility functions like this should be defined outside of the class (or as a static function in pure OOP languages). The function doesn't require a class instance to be defined, so being inside a class just adds extra mental overhead/context.

olzraiti: @MLumme is on vacation so I took over on making the code changes & approved

olzraiti: The class has also other utils as class methods, so for consistency I didn't want to introduce a "new scope", so I put the function inside the class method that needs the util

olzraiti (approved):

PR #410, opened 2023-05-23
fix breeding index annotation bug

Blodir:
https://www.pivotaltracker.com/story/show/185230510


rpulkka (approved):
The fix creates a tag type "check" and uses it in the checkTypeTag -function by default instead of using a parameter.

I like that you used the checkTypeTagPasses variable for readability. In my opinion, also the rest of the conditions in disableSend (which already existed before this PR) should be refactored sometime later, because the conditions in total are quite difficult to read at the moment. The conditions could be equated to readable variables like you did. But this is beyond the scope of this PR, so let's write a new refactoring task on that to Pivotal.

Good work, approving!

PR #411, opened 2023-05-23
Polygon search on firefox android fix

Blodir:
https://www.pivotaltracker.com/story/show/178601598


aorin (approved):
I tried to test it with android firefox but I couldn't reproduce it in beta. It's quite strange but the fix is ok

PR #413, opened 2023-05-23
Document permission service change

aorin:
https://185177956.dev.laji.fi
https://www.pivotaltracker.com/n/projects/2346653

Previously the document permission service used both form ID and collection ID to determine if the user is a form admin but all the documents don't have a form ID and the form check is unnecessary so now it uses only the collection ID.


olzraiti (approved):
LGTM

PR #420, opened 2023-08-14
Fix deleting local document [#184661048]

olzraiti:
https://www.pivotaltracker.com/story/show/185412289
https://185412289.dev.laji.fi/vihko

Old implementation tried to always delete the document through the API. It failed for local documents since they don't exist, resulting in a "hups" and the local doc not being removed.

I fixed the logic to first check if the doc is local one, and accordingly remove it from the local or remote source.

### DIFF
--- a/projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
+++ b/projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
@@ -51,13 +51,13 @@ export class DocumentService {
   }
 
   deleteDocument(id: string) {
-    return this.documentApi.delete(id, this.userService.getToken()).pipe(
-      mergeMap(() => this.userService.user$),
-      tap(person => {
+    const isTmpDoc = id.match(/^T:\d+$/);
+    const del$ = isTmpDoc
+      ? this.userService.user$.pipe(tap(person => {
         this.documentStorage.removeItem(id, person);
-        delete this.cache[this.getCacheKey(id)];
-      })
-    );
+      }))
+      : this.documentApi.delete(id, this.userService.getToken());
+    return del$.pipe(tap(() => delete this.cache[this.getCacheKey(id)]));
   }
 
   saveTemplate(templateData: TemplateForm): Observable<Document> {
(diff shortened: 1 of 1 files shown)


projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
```diff
@@ -51,13 +51,13 @@ export class DocumentService {
   }
 
   deleteDocument(id: string) {
-    return this.documentApi.delete(id, this.userService.getToken()).pipe(
-      mergeMap(() => this.userService.user$),
-      tap(person => {
+    const isTmpDoc = id.match(/^T:\d+$/);
+    const del$ = isTmpDoc
+      ? this.userService.user$.pipe(tap(person => {
```
aorin, line 56: I think this should have `user$.pipe(take(1))` because we only want the current user

projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
```diff
@@ -51,13 +51,13 @@ export class DocumentService {
   }
 
   deleteDocument(id: string) {
-    return this.documentApi.delete(id, this.userService.getToken()).pipe(
-      mergeMap(() => this.userService.user$),
-      tap(person => {
+    const isTmpDoc = id.match(/^T:\d+$/);
```
aorin, line 54: This could use the method `FormService.isTmpId`

aorin (changes requested):

projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
olzraiti: We do that in numerous places, but I'm pretty sure it's unnecessary. Logging in/out reloads the page and thus the `user$` pipe will always return just a single user?

projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
olzraiti: Thanks, applied

projects/laji/src/app/shared-modules/own-submissions/service/document.service.ts
olzraiti: OK I tested that logging out doesn't reload the page so better add the `take(1)` so that logging out won't do anything weird

aorin (approved):
Looks good

PR #423, opened 2023-08-16
Bird Atlas update

Blodir:
Includes several bird atlas related tasks that are already in production (https://tulokset.lintuatlas.fi/)


MLumme (approved):
LGTM, although due to its size can't promise some oddities won't slip trough.

PR #424, opened 2023-08-16
Obs viewer fails if geometry missing #185412397

MLumme:
Task here https://www.pivotaltracker.com/story/show/185412397, there were couple of places where code assumed wrongly that there would be coordinates.


olzraiti (approved):
LGTM

PR #425, opened 2023-08-16
Disable scrolling on observation search numeric fields #185645030

MLumme:
Task here https://www.pivotaltracker.com/story/show/185645030, added a directive that can be added to fields to disable scroll events.


olzraiti (approved):
LGTM

PR #427, opened 2023-08-16
Punainenkirja feedback form broken #184670721

MLumme:
Task here https://www.pivotaltracker.com/story/show/184670721, problem is that in red book there is no user, and RXJS does not execute observables and pipes if that observable is an empty value, like in userless case here, so now feedback form has an check based on presence of user persontoken to chose how it does the feedback sending.

### DIFF
--- a/projects/laji/src/app/shared/feedback/feedback.component.ts
+++ b/projects/laji/src/app/shared/feedback/feedback.component.ts
@@ -61,33 +61,43 @@ export class FeedbackComponent {
       return;
     }
     const meta = this.getMeta();
-    this.userService.user$.pipe(
-      take(1),
-      switchMap(user => this.lajiApi.post(
+    const onSendEnd = {
+      next: () => {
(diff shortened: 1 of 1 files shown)


MLumme:
> The logic seems fine, but I think it'd help with readability if it was written in a single rxjs stream like so:
> 
> ```
> (!this.userService.getToken()
>  ? of(undefined)
>  : this.userService.user$.pipe(
> 	 take(1),
> 	 switchMap(user => send(user))
> 	 )
> ).pipe(<logic from send() here>)
> .subscribe(<logic from end here>)
> ```
> 
> It would be read in a more chronological order this way, instead of having to jump to the `send()` and `onSendEnd()` when reading the code.

Yeah, I agree that it seems cleaner that way, WILCO.

MLumme:
@olzraiti Implemented your recommended changes.

olzraiti (approved):
Thanks :+1:

PR #428, opened 2023-08-16
Column sorting usability 180887524

MLumme:
Task here https://www.pivotaltracker.com/story/show/180887524, wrote logic to put sorting changes reversed into the sorting array.

### DIFF
--- a/projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
+++ b/projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
@@ -300,13 +300,13 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnChanges, OnD
   //underlying ngx-datatable uses external sorting
   private updateSorts(event) {
     if (!event.newValue) {
-      return this.sorts.filter(sort => sort.prop !== event.column.name);
+      return this.sorts.filter(sort => sort.prop !== event.column.prop);
     }
 
-    const idx = this.sorts.findIndex(sort => sort.prop === event.column.name);
+    const idx = this.sorts.findIndex(sort => sort.prop === event.column.prop);
     if (idx > -1) {
       return this.sorts.map(sort => {
-        if (sort.prop === event.column.name) {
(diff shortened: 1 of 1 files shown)


olzraiti (changes requested):
At least the following bug:

1. goto https://180887524.dev.laji.fi/observation/list
2. click taxon column
-> doesn't sort

~I'm assuming that's something about the component having `OnPush` as the change detection method (which is good), and the `updateSorts` returns `this.sorts`, which is immutably edited, but it's still the same array instance and thus the change detection fails even when it's manually called to detect changes. Making the return points of `updateSorts()` to always make a copy of `this.sorts` before returning could fix this?~

EDIT: the above was wrong - `updateSorts` returns always a new instance so the hypothesis isn't right.

projects/laji/src/app/shared-modules/datatable/datatable/datatable.component.ts
```diff
@@ -287,12 +287,43 @@ export class DatatableComponent implements AfterViewInit, OnInit, OnChanges, OnD
   }
 
   onSort(event) {
-    this.sorts = event.sorts;
-    this.sortRows(event.sorts);
+    this.sorts = this.updateSorts(event);
+    this.sortRows(this.sorts);
     this.changeDetectorRef.detectChanges();
+    event.sorts = this.sorts;
```
olzraiti, line 293: Is this necessary? Altering function parameters in mutable fashion is shady business.

MLumme: Actually not, I should've put a new object into event emitter that is the new sort and rest from destructuring the event, I'll fix it. And I'll try to figure out what's going on with that bug.

MLumme: It's fixed now, I had set wrong column property as sort prop, which worked most of the time by accident but not for taxon.

olzraiti (approved):
Thanks! :+1:

PR #432, opened 2023-08-16
Add technical news to all pages #185049817

MLumme:
Task here https://www.pivotaltracker.com/story/show/185049817, probably not on all pages, but then news should now show up on all taxon pages and cms-pages.

### DIFF
--- a/projects/laji/src/app/+information/information.component.html
+++ b/projects/laji/src/app/+information/information.component.html
@@ -1,8 +1,10 @@
+TEST
 <ng-container *ngIf="information$ | async; let information">
   <lu-sidebar position="right" [staticWidth]="280" [displayNav]="!!information.children">
     <main class="d-flex">
       <div class="d-flex justify-center flex-grow">
         <div class="mt-4 mb-8 mx-4 d-flex flex-col flex-nowrap width-0 information-container">
+          <laji-technical-news></laji-technical-news>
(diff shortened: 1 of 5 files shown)


projects/laji/src/app/+information/information.module.ts
```diff
@@ -4,9 +4,9 @@ import { SharedModule } from '../shared/shared.module';
 import { routing } from './information.routes';
 import { InformationComponent } from './information.component';
 import { LajiUiModule } from 'projects/laji-ui/src/public-api';
-
+import { TechnicalNewsModule } from '../shared-modules/technical-news/technical-news.module';
```
aorin, line 7: Not a big issue but it looks a little weird that there is no space between the lines `import ...` and `@NgModule ...`

projects/laji/src/app/+information/information.component.html
```diff
@@ -1,8 +1,10 @@
+TEST
```
aorin, line 1: This should be removed

aorin (changes requested):

projects/laji/src/app/+information/information.component.html
MLumme: Both done.

aorin (approved):
:+1:

PR #433, opened 2023-08-16
Fix taxa shown on bumblebee census results #184626236

MLumme:
TAsk here https://www.pivotaltracker.com/story/show/184626236, problem was that in the tables scientific name was used to identify taxons, which informal taxon groups do not have, so they were identified as undefined and in the table latter informal groups would just overwrite earlier ones. using taxon id now as identifier.


aorin (approved):
Looks ok

PR #436, opened 2023-08-17
Selecting year range as time #184710661

MLumme:
Task here https://www.pivotaltracker.com/story/show/184710661, also added the handler for keypress' to catch enter-keys to set the value.

### DIFF
--- a/projects/laji/src/app/shared/datepicker/datepicker.component.ts
+++ b/projects/laji/src/app/shared/datepicker/datepicker.component.ts
@@ -108,7 +108,7 @@ export class DatePickerComponent implements ControlValueAccessor {
         return this.updateValue(momentValue.format(FORMAT));
       }
 
-      this.validDate = false
+      this.validDate = false;
       return this.onInputValueChange(this.viewValue);
     }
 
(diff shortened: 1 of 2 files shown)


projects/laji/src/app/shared/datepicker/datepicker.component.ts
```diff
@@ -90,27 +90,32 @@ export class DatePickerComponent implements ControlValueAccessor {
...
-  onInputValueChange(viewFormatValue: string) {
-    const viewFormMoment = moment(viewFormatValue, VIEW_FORMAT, true);
-
-    if (this.validDate && viewFormatValue && !viewFormMoment.isValid()) {
-      this.validDate = false;
-      return this.onInputValueChange(this.viewValue); // Restore prev value that is valid.
-    } else {
-      this.validDate = true;
+  keyEvent(e, value) {
+    if (e.keyCode === 13) {
```
Blodir, line 94: Use keycodes from angular cdk https://github.com/angular/components/blob/main/src/cdk/keycodes/keycodes.ts
```
import { ENTER } from '@angular/cdk/keycodes';
```

Blodir (changes requested):

projects/laji/src/app/shared/datepicker/datepicker.component.ts
MLumme: Done.

Blodir (approved):

PR #438, opened 2023-08-29
SSR fixes

aorin:
Fixes two SSR issues

- `datepicker.component` threw warnings when constructing an invalid date
- `LajiFormBuilder` import threw errors with SSR


olzraiti (approved):
Thanks, looks good.

PR #439, opened 2023-09-21
Habitat filter or is broken 184872045

MLumme:
Task here https://www.pivotaltracker.com/story/show/184872045.

### DIFF
--- a/projects/laji/src/app/+observation/download/observation-download.component.ts
+++ b/projects/laji/src/app/+observation/download/observation-download.component.ts
@@ -185,7 +185,7 @@ export class ObservationDownloadComponent implements OnDestroy {
 
   updateCount() {
     this.observationDataService.getData(this._originalQuery).pipe(
-      map(data => data.secureCount)
+      map(data => data.securedCount)
     ).subscribe(total => {
       this.privateCount = total;
       this.cd.markForCheck();
--- a/projects/laji/src/app/+observation/download/observation-download.component.ts
+++ b/projects/laji/src/app/+observation/download/observation-download.component.ts
@@ -185,13 +185,7 @@ export class ObservationDownloadComponent implements OnDestroy {
 
   updateCount() {
     this.observationDataService.getData(this._originalQuery).pipe(
-      map(data => data.private.total),
-      catchError(() => this.warehouseService.warehouseQueryCountGet({
-        ...this.query,
-        secured: true
-      }).pipe(
-        map(result => result.total)
-      ))
+      map(data => data.secureCount)
     ).subscribe(total => {
       this.privateCount = total;
       this.cd.markForCheck();
--- a/projects/laji/src/app/+observation/observation-data.service.ts
+++ b/projects/laji/src/app/+observation/observation-data.service.ts
@@ -6,9 +6,10 @@ import { WarehouseQueryInterface } from '../shared/model/WarehouseQueryInterface
 import { SearchQueryService } from './search-query.service';
 import { WarehouseApi } from '../shared/api/WarehouseApi';
 import { PlatformService } from '../root/platform.service';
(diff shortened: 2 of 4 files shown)


projects/laji/src/app/+observation/observation-data.service.ts
```diff
@@ -1,26 +1,16 @@
...
-  };
-  species: {
-    total: number;
-  };
-  private: {
-    total: number;
-  };
+  count: number;
+  speciesCount: number;
+  securedCount: number;
```
olzraiti, line 13: Good job on simplifying the datamodel. Could be get a more descriptive name for `count` as well? Count of what, `unitsCount`?

projects/laji/src/app/+observation/download/observation-download.component.ts
```diff
@@ -185,13 +185,7 @@ export class ObservationDownloadComponent implements OnDestroy {
...
-      map(data => data.private.total),
-      catchError(() => this.warehouseService.warehouseQueryCountGet({
-        ...this.query,
-        secured: true
-      }).pipe(
-        map(result => result.total)
-      ))
+      map(data => data.securedCount)
     ).subscribe(total => {
       this.privateCount = total;
```
olzraiti, line 190: `privateCount` is just a copy of `secureCount`. `privateCount` could be removed?

projects/laji/src/app/+observation/observation-data.service.ts
```diff
@@ -31,105 +21,55 @@ const overrideType = {
...
+    private warehouseService: WarehouseApi
   ) { }
 
   getData(query: WarehouseQueryInterface): Observable<IObservationData> {
     if (this.platformService.isServer) {
       return of(this.empty());
     }
+    const newQuery = JSON.stringify(query);
+
+    if (newQuery === this.lastQuery) {
```
olzraiti, line 39: This will give false negatives, if the "order" of properties changes (`{"a":1,"b":2} !== {"b":2,"a":1}`). You could use for example the `deep-equal` package for checking the similarity.

olzraiti (changes requested):
Left some comments.

Also, there's a bug:

1. goto https://184872045.dev.laji.fi/observation/map?target=MX.40573&time=2023-09-20%2F
-> "osumaa" & "lajista" spin forever

projects/laji/src/app/+observation/observation-data.service.ts
MLumme: Done with deep-equal.

MLumme:
> Left some comments.
> 
> Also, there's a bug:
> 
>     1. goto https://184872045.dev.laji.fi/observation/map?target=MX.40573&time=2023-09-20%2F
>        -> "osumaa" & "lajista" spin forever

Fixed the bug, there was a bit of an problem with javascritp falsy values when the counts were zero.

projects/laji/src/app/+observation/download/observation-download.component.ts
olzraiti: Yeah it'd require some refactoring to get this working with observables instead of the class instance variables... It's ok as it is :+1:

projects/laji/src/app/+observation/observation-data.service.ts
olzraiti: Thanks. One more thing - apparently these can also be `null`. Could you add that to the typings? That should increase safety. I wonder how the compiler doesn't complaing about setting these to `null` with the `empty()` method... :thinking:

olzraiti: Well, never mind, that's out of scope of this PR since it was already that way.

olzraiti (approved):
Thanks!

PR #441, opened 2023-09-21
Line transect page makes duplicate requests 182152439

MLumme:
Task here https://www.pivotaltracker.com/story/show/182152439


rpulkka (approved):
Approving! PR adds RxJS shareReplay to prevent double-triggering the request to /named-places.

PR #448, opened 2023-11-07
Dashboard link [#186219898]

rpulkka:
https://130387094.dev.laji.fi/
https://www.pivotaltracker.com/n/projects/2346653/stories/186219898

NOTE: The feature branch address has a different code, because it is a copy of this branch, which I created because CI had issues creating the feature branch page when this was pushed to Git.

This PR contains a link to the Dashboard. It passes the user locale i.e. opens the Dashboard in English if laji-front has English selected. The link also shows the count of observations that were added in the last 7 days.


MLumme (approved):
LGTM

PR #449, opened 2023-11-07
Change dataset link to a local link [#186228805]

rpulkka:
https://186228805.dev.laji.fi/save-observations
https://www.pivotaltracker.com/n/projects/2346653/stories/186228805

The dataset-page link pointed to production even in the dev-laji version. Now it is a local link that points to dev while using dev-laji.


rpulkka:
Closed because this change could be implemented just by changing the link translation to Crowdin.

PR #451, opened 2023-11-07
Use plausible in prod only

Blodir:
Production and dev environments now use different `index.html` files and plausible is only included in production.

https://www.pivotaltracker.com/story/show/185842878


MLumme (approved):
LGTM

PR #455, opened 2023-11-08
Allow mapping to empty value 184272933

MLumme:
Task https://www.pivotaltracker.com/story/show/184272933, testable here https://184272933.dev.laji.fi/.


rpulkka (changes requested):
Potential issue: When "Havainnointiminuutit - Keruutapahtuma" field is 0, and the linking columns -stage links it to empty, the user can still press arrow up and arrow down, resulting the value to go "empty -> 1", "1 -> 0" even though the value should be a positive integer, and user can still continue.

rpulkka (approved):
Approving. Further import validation issues will be fixed in an upcoming Pivotal task.

PR #456, opened 2023-11-09
Image tab mode persists [#186078537]

rpulkka:
https://186078537.dev.laji.fi/observation/images
https://www.pivotaltracker.com/n/projects/2346653/stories/186078537

Now the image tab's view mode (large images or small images) persists after changing the paginator page. This is done by emitting the view state to parent. Also ViewType is implemented in this PR as a refactor.


aorin (approved):
:+1:

PR #457, opened 2023-11-09
BSG changes

aorin:
This contains various changes to BSG (https://kerttu-global-dev.laji.fi/)

- support for bat and insect recordings
- recordings are preloaded to cache in the Identify Recordings section
- adds "My annotations" tab to the Identify Recordings section
- fixes some issues related to the Angular update

None of the changes affect any other project.


Blodir (approved):
I added a few comments, but I don't think they are showstoppers or anything.

Also I saw lots of the C# naming convention of starting interface names with I, this not very typescripty/angulary practice. I think I've already commented on this though, and the project already has lots of this naming convention so I guess it makes sense to stick to it for this project. In laji project I hope we can get rid of it though :)

PR #458, opened 2023-11-09
Fix modal doesn't touch body overflow if it's not been hijacked yet

olzraiti:
Currently the modal hides itself on construction. This can cause it to remove body's overflow. This PR makes the body overflow control more fine-grained: modal construction hides the main element, but doesn't touch body overflow until the modal is actually shown. Then, the modal "hijacks" the body overflow, and "releases" it when it hides.

### DIFF
--- a/projects/laji-ui/src/lib/modal/modal/modal.component.ts
+++ b/projects/laji-ui/src/lib/modal/modal/modal.component.ts
@@ -94,7 +94,7 @@ export class ModalComponent implements OnDestroy {
       return false;
     }
     this.hideElement();
-    this.restoreBodyOverflow();
+    this.releaseBodyOverflow();
     this.isShown = false;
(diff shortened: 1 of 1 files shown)


projects/laji-ui/src/lib/modal/modal/modal.component.ts
```diff
@@ -82,10 +82,8 @@ export class ModalComponent implements OnDestroy {
     if (this.isShown) {
       return;
     }
-    this.originalBodyOverflow = this.document.body.style.overflow;
     this.renderer.setStyle(this.elementRef.nativeElement, 'display', 'block');
```
aorin, line 85: For consistency this could be refactored to function `showElement` (as there is the function `hideElement`)

aorin (approved):
:+1:

PR #461, opened 2023-11-14
Bring back redis cache to ssr

olzraiti:
﻿Brings back the Redis cache to SSR, that was removed on `ng@16` update (the SSR server was automatically updated by ng update, and the cache solution was implemented to the previously autogenerated server).

This solution is based on the old SSR server that was customized: https://github.com/luomus/laji/blob/e8d99f820f17830a4f6c4ccc2eb3eb4eb3c7a20e/server.ts

To make future updates easier, I moved the cache logic to a separate file that exports an Express middleware that does the caching. Changes to the autogenerated `server.ts` are as minimal as possible: a single middleware is added to the render endpoint.

Since I wanted minimal changes to the `server.ts`, the middleware has to do some hacking around to be able to perform - it hijacks the `response.render()` method. This is carefully documented in the code.

I added a wiki note about this to our [Angular update guide](https://wiki.helsinki.fi/display/luomusict/Updating+laji.fi+dependencies)

Logic explained:

It's an optimistic cache that clears itself after `CACHE_TIME` (30min) has expired. It updates the cached data on cache hit if `CACHE_UPDATE` (30s) has passed since the last hit.


MLumme (approved):
LGTM, at least based on my bit rusty express.

PR #462, opened 2023-11-15
Fix window references

olzraiti:
Found during SSR testing that these references to `window` made the SSR crash.

The `window` should be accessed through the platform browser, because it's not in node API.


Blodir (approved):
-- looks good

PR #463, opened 2023-11-15
Modal fixes

olzraiti:
Check the commit messages for description for what is fixed (click them open, they are multi line messages)

### DIFF
--- a/projects/laji-ui/src/lib/_vars.scss
+++ b/projects/laji-ui/src/lib/_vars.scss
@@ -9,6 +9,7 @@ $sp-offset: 0px !default;
 */
 
 $breakpoints: (
+  "xl": 1600px,
   "lg": 1200px,
   "md": 992px,
   "sm": 768px,
--- a/projects/laji-ui/src/lib/modal/modal.service.ts
+++ b/projects/laji-ui/src/lib/modal/modal.service.ts
@@ -37,7 +37,7 @@ export class ModalService {
   show<T>(componentClassOrTemplateRef: Type<T> | TemplateRef<T>, options?: ModalOptions<T>): ModalRef<T> {
     const modalComponent = this.showModalInBody(options);
     const contentNode = modalComponent.instance.getContentNode();
-    const contentHostNode = document.createElement('div');
+    const contentHostNode = this.document.createElement('div');
     this.renderer.appendChild(contentNode, contentHostNode);
     this.modal = modalComponent.instance as any;
     this.modalComponent = modalComponent;
@@ -55,7 +55,7 @@ export class ModalService {
   private showModalInBody<T>(options: ModalOptions<T> = {}) {
     const modalComponent = createComponent(ModalComponent, { environmentInjector: this.appRef.injector });
     this.appRef.attachView(modalComponent.hostView);
-    this.renderer.appendChild(document.body, modalComponent.location.nativeElement);
+    this.renderer.appendChild(this.document.body, modalComponent.location.nativeElement);
     Object.keys(options).forEach(option => {
       modalComponent.instance[option] = options[option];
     });
@@ -100,7 +100,7 @@ export class ModalService {
     this.modalComponent = undefined;
     this.content.destroy();
     this.content = undefined;
-    this.renderer.removeChild(document.body, (this.modal as any).elementRef.nativeElement);
+    this.renderer.removeChild(this.document.body, (this.modal as any).elementRef.nativeElement);
     this.modal = undefined;
   }
 }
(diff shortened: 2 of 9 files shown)


projects/laji-ui/src/lib/_vars.scss
```diff
@@ -9,6 +9,7 @@ $sp-offset: 0px !default;
 */
 
 $breakpoints: (
+  "xl": 1600px,
```
Blodir, line 12: Could this use `$xl-width` etc.?

projects/laji-ui/src/lib/modal/modal/modal.component.scss
```diff
@@ -74,14 +86,26 @@
...
+      left: calc(50% + $lg-width / 2 + $padding);
     }
-      > .lu-modal-close-button {
-        left: calc(50% + 455px);
-      }
 	}
 }
 
+
+@media (min-width: 1660px) {
```
Blodir, line 98: use the breakpoint variable here? `$xl-width`

Blodir (changes requested):

projects/laji-ui/src/lib/_vars.scss
olzraiti: For every other var (sm, md, lg) those numbers are different between breakpoints and the widths

projects/laji-ui/src/lib/modal/modal.service.ts
```diff
@@ -58,7 +61,8 @@ export class ModalService {
     });
     modalComponent.instance.show();
     modalComponent.changeDetectorRef.detectChanges();
-    modalComponent.instance.onHide.subscribe(this.destroyOnHide.bind(this));
+    this.modalHideSub?.unsubscribe();
```
aorin, line 64: Now it doesn't call the `destroyOnHide` function for all modals and some modals are not removed. I tested opening two modals and hiding them (in here http://localhost:3000/observation/list?informalTaxonGroupId=MVL.26) and it left the first modal in the html (it had `display:none` but it's not optimal that it's are not removed).

aorin (changes requested):

projects/laji-ui/src/lib/modal/modal/modal.component.scss
Blodir: Right, sorry. Anyways these sort of feel like magic values when they should be related to breakpoints. Could maybe do something like `calc(map-get($breakpoints, "xl") + 60px)` (not sure if that syntax works but you get the idea)

Blodir: Not sure why the breakpoints should have exceptions here though

olzraiti: Hmm... The breakpoints are weird numbers like 576, 768, 992, 1660. There's no pattern.

Modal widths are numbers like 300, 600, 900, 1600. There's a pattern (ok there's an exception but it's kinda related to the pattern still).

I don't see a pattern between the breakpoints and the widths. If we'd change some of the breakpoint values (I don't think we ever will), the width pattern would break.

Also I don't think they should be relative anyways - think about the small width. If the small breakpoint is changed, the width of the small modal would change for all widths and that's not necessarily desired.

projects/laji-ui/src/lib/modal/modal.service.ts
olzraiti: Well, that escalated somewhat... There was a design flaw that I was aware of but didn't want to spend effort on fixing since it was already in the bootstrap API or how we used it at least. See commit message for details.

projects/laji-ui/src/lib/modal/modal/modal.component.scss
Blodir: I probably should have read the commit more carefully... But it just feels weird to have exceptions in media queries. Like the entire purpose of the breakpoints is media queries like this :D I'm not sure if it actually matters though (does it matter that things wrap etc at different breakpoints? idk). I think the pr can be approved as is.

projects/laji-ui/src/lib/modal/modal.service.ts
aorin: I think it's good idea to remove the references to the latest modal in the modal service. I checked how the hiding inside the component worked in the ngx-bootstrap and it works so that the component can get a reference to its "own" modal by injecting `modalRef` (the modal service provides it when creating the component) and then the component can call `modalRef.hide()`. It would be a nice-to-have feature but not necessary so I think this is good to go

aorin (approved):

Blodir (approved):

PR #464, opened 2023-11-17
Fix webgl memory leak

Blodir:
https://luomus-ict.slack.com/archives/CQEQSQCP7/p1700214321673549


olzraiti (approved):
no idea what's going on here, approving

PR #465, opened 2023-11-22
fix laji image modal overlay for images

olzraiti:
Broken by `53b3bbe7f4417c11e09e8bbf315cf99d024093d6`.

`MM.image` doesn't have `mediaType` property in our schema, so idk whether it's supposed to be there. Anyways, it wasn't there for images at least.

Repro: image modal broken on https://beta.laji.fi/taxon/MX.53697/identification and https://beta.laji.fi/taxon/MX.53697/images


Blodir (approved):
Not sure why `mediaType` is not in schema. Here's an example with `mediaType` present:

https://beta.laji.fi/api/warehouse/query/single?documentId=http://id.luomus.fi/KX.1325

PR #467, opened 2023-11-23
ESLint memory increase [#186597346]

rpulkka:
https://www.pivotaltracker.com/n/projects/2346653/stories/186597346

Added JS Heap Size increase to ESLint scripts and updated the project contributors.


projects/laji/src/app/+taxonomy/taxon/taxon-tree/tree/tree.component.html
```diff
@@ -3,7 +3,7 @@
     <ng-container *ngFor="let node of tree.parentNodeList; let last = last">
       <div *ngIf="!node.state.isSkipped || node.value.id === activeId">
         <laji-spinner *ngIf="last && node.state.loadingCount > 0"></laji-spinner>
-        <ng-container *ngTemplateOutlet="labelTpl; context: {'value': node.value}"></ng-container>
+        <ng-content /><ng-container *ngTemplateOutlet="labelTpl; context: {'value': node.value}"></ng-container>
```
Blodir, line 6: This is unnecessarily hacky. We weren't able to reproduce the issue, but it appears that this is what is happening:
- `taxon-tree.component` passes the label template as content projection
- `tree.component.ts`: `@ContentChild` gets a reference to that template, but this behavior appears to be buggy (maybe it gets the template only after the tree has been rendered so the tree doesn't have access to it on the first render? idk)
- adding `<ng-content/>` actually does the content projection. Perhaps this ensures that the template is indeed available? Note that even if this works we are getting the reference to the template twice, and it's really hard to tell if everything is working correctly since that would require some deeper knowledge of angular internal magic

Better solution: Don't use content projection on `ng-template`s. Instead pass a reference to the template as a `@Input`. I created a new branch for this (avoid cluttering commit history, and the package.json change had issues too).  https://github.com/luomus/laji/pull/471

Blodir (changes requested):

rpulkka:
Can be discarded, as  #471 includes a better fix.

olzraiti:
@rpulkka if you're up for it, we could turn this into lint JS memory increase PR. We'd need to find a way to make it work cross-env (for example see "third option" in the answer of https://stackoverflow.com/questions/56982005/where-do-i-set-node-options-max-old-space-size-2048)

rpulkka:
@olzraiti OK, I wrote a Pivotal task so that I remember to do it: https://www.pivotaltracker.com/n/projects/2346653/stories/186597346

rpulkka:
@Blodir Can you test if the ESLint scripts work on Windows with the cross-env module, which I added to the project? If yes, can you also approve the Pivotal story: https://www.pivotaltracker.com/n/projects/2346653/stories/186597346

Blodir (approved):

PR #470, opened 2023-11-30
Hydration fixes

Blodir:
- remove the ssr disable directive that broke hydration for buttons/inputs
- fix an instance of invalid html (forbidden elements inside a `p` element) that broke hydration


olzraiti (approved):
Merging to beta. If we don't enable hydration on prod yet on the next rel, we might have to push some patch to get the ssr disable back.

PR #473, opened 2023-12-11
Migrate user e2e tests

Blodir:
I didn't migrate `user.po.ts`. Stuff in there should be implemented when/if they are needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ npm start
 ```
 
 ## Running end-to-end tests
-1. Get a permanent person token via https://fmnh-ws-test.it.helsinki.fi/laji-auth/login?target=KE.389&redirectMethod=GET&locale=fi&next=&offerPermanent=true
+1. Get a permanent test user person token from another developer (the tests assume that this user is being used)
 2. Create a `.env` file at the root of the repository
 3. Insert the token `PERSON_TOKEN=<your_token>`
 4. run `npm start`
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -25,22 +25,12 @@ npm start
...
-
 ## Running end-to-end tests
-
-Run `ng e2e` to execute the end-to-end tests via [Protractor](http://www.protractortest.org/).
-Before running the tests make sure you are serving the app via `ng serve`.
+1. Get a permanent person token via https://fmnh-ws-test.it.helsinki.fi/laji-auth/login?target=KE.389&redirectMethod=GET&locale=fi&next=&offerPermanent=true
+2. Create a `.env` file at the root of the repository
+3. Insert the token `PERSON_TOKEN=<your_token>`
+4. run `npm start`
+5. run `npx playwright test --ui`
```
olzraiti, line 33: In practice our tests assume that they are ran against a certain person token, and it takes quite a lot of configuring to create an identical person setup (correct form permission rights etc)... Maybe we could mention that you need to have the secret token so people wouldn't try to user e.g. their personal account for the tests? "Ask a collegue for the token"?

olzraiti (changes requested):

README.md
Blodir: Pretty suspicious practice that the tests only work for a specific user :/ but let's just go with that for now...

olzraiti (approved):

PR #475, opened 2023-12-18
Refactor media-filter and add model filter

Blodir:
https://www.pivotaltracker.com/story/show/183872808


projects/laji/src/app/+observation/form/media-filter/media-filter.component.ts
```diff
@@ -0,0 +1,65 @@
+import { ChangeDetectionStrategy, Component, EventEmitter, Input, Output } from '@angular/core';
+import { WarehouseQueryInterface } from '../../../shared/model/WarehouseQueryInterface';
+
+const mediaGroupProps = ['hasUnitImages', 'hasUnitAudio', 'hasUnitModel'] as const;
```
aorin, line 4: This could have a type that is linked with `WarehouseQueryInterface`
```
type MediaGroupProp = keyof Pick<WarehouseQueryInterface, 'hasUnitImages'|'hasUnitAudio'|'hasUnitModel'>;
const mediaGroupProps: MediaGroupProp[] = ...
```
Then it would throw an error if some of the props had a typo (I guess this doesn't do it?). But not that important thing

aorin (approved):
Looks good, the component looks much clearer now

PR #477, opened 2023-12-28
Add dockerized npm-run

olzraiti:
This allows for example building the app regardless of the environment. Node updates should be a lesser problem.

Usage: `npm run docker:npm-run build:ssr:beta`

Deletes also old docker files, I don't think anybody is using docker for local development.


rpulkka (approved):
OK! Tested with these, it creates a build successfully:
`sudo npm run docker:npm-run build:ssr:beta`
`sudo docker cp $(sudo docker ps -alq):/app/dist/browser browserz`
Dockerfile looks fine also.

PR #479, opened 2023-12-30
Fix eslint not finding tsconfig on some systems

Blodir:
Running neovim on windows eslint was unable to find tsconfig (it thought that the root directory was at the location of the current file...). Unfortunately I was not able to find a fix that doesn't touch project files, but I don't think this fix should break anything.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,4 +1,4 @@
-{
+module.exports = {
   "root": true,
   "ignorePatterns": [
     "projects/laji/src/test.ts"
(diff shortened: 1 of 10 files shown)


.eslintrc.js
```diff
@@ -9,6 +9,7 @@
         "*.ts"
       ],
       "parserOptions": {
+        tsconfigRootDir: __dirname,
```
aorin, line 12: All other properties have quotes so this could also have them (`"tsconfigRootDir"`)

aorin (changes requested):
Tested and works ok

.eslintrc.js
Blodir: Ideally none of the property keys would have quotes in a js file (this is a fragment of changing the file from json to js) but I'll just add the quotes for now

aorin (approved):

PR #481, opened 2024-01-02
Playwright theme migration

aorin:
This could be checked after the [project-form PR](https://github.com/luomus/laji/pull/474) has been merged to the `playwright` branch since this branch is based on the `playwright-project-form` branch.

### DIFF
--- a/projects/laji/test/+theme/genetic-resource/genetic-resource.po.ts
+++ b/projects/laji/test/+theme/genetic-resource/genetic-resource.po.ts
@@ -1,0 +1,26 @@
+import { Page } from '@playwright/test';
+
+export class GeneticResourcePage {
+  public $spinners = this.page.locator('laji-observation-result').locator('.spinner');
+
+  public hasNonGQLApiErrors = false;
+
+  constructor(
+    private page: Page
+  ) {
+    this.page.on('response', resp => {
+      if (resp.url().match(/.*api\/(?!graphql).*/) && [400, 404, 500].includes(resp.status())) {
+        this.hasNonGQLApiErrors = true;
+      }
+    });
+  }
+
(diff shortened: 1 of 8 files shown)


projects/laji/test/+theme/genetic-resource/genetic-resource.po.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class GeneticResourcePage {
+  public $spinners = this.page.locator('laji-observation-result').locator('.spinner');
+
+  public hasNonGQLApiErrors = false;
+
+  constructor(
+    private page: Page
+  ) {
+    this.page.on('response', resp => {
```
Blodir, line 11: Maybe this could be extracted as a helper function instead? It seems a bit strange that the user of GeneticResourcePage must be aware of when the value of `hasNonGQLApiErrors` changes. So basically they have to read the implementation to use the class.

Blodir: could be something like `export const responseHasNonGQLApiErrors = (response): boolean => ...`

and then this.page.on... would be called in genetic-resource.spec.ts

Blodir (changes requested):
Sorry for taking so long getting to the review. The tests pass for me.

projects/laji/test/+theme/genetic-resource/genetic-resource.po.ts
Blodir: Actually it's probably fine since the `hasNonGQLApiErrors` is always up to date 🤔 However, resp should be typed with `import { Response } from '@playwright/test'`

Blodir (approved):

PR #482, opened 2024-01-15
Stricter TS config continued

olzraiti:
https://241201.dev.laji.fi

Applies strict ts config for some parts of the project.

### DIFF
--- a/projects/laji/src/app/shared/navbar/notifications/notifications.component.ts
+++ b/projects/laji/src/app/shared/navbar/notifications/notifications.component.ts
@@ -1,7 +1,4 @@
-import {
-  Component, ChangeDetectionStrategy, OnDestroy, OnInit, Output,
-  ChangeDetectorRef, Renderer2, ElementRef, AfterViewInit, EventEmitter, ViewChild
-} from '@angular/core';
+import { Component, ChangeDetectionStrategy, OnDestroy, OnInit, ChangeDetectorRef, ViewChild } from '@angular/core';
 import { of, Subject } from 'rxjs';
 import { switchMap, takeUntil, filter, tap, map } from 'rxjs/operators';
 import { NotificationsFacade } from './notifications.facade';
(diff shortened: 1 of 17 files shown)


projects/laji/src/app/shared/navbar/notifications/notifications.component.ts
```diff
@@ -20,10 +17,10 @@ import { DialogService } from '../../service/dialog.service';
 export class NotificationsComponent implements OnInit, OnDestroy {
   unsubscribe$ = new Subject<void>();
 
-  notificationSource: NotificationDataSource;
+  notificationSource!: NotificationDataSource;
 
-  @ViewChild(CdkVirtualScrollViewport, {static: true}) virtualScroll: CdkVirtualScrollViewport;
-  loading: boolean;
+  @ViewChild(CdkVirtualScrollViewport, {static: true}) virtualScroll!: CdkVirtualScrollViewport;
+  loading!: boolean;
```
aorin, line 23: I think using `!` should be avoided when it's possible (with the exception of static `ViewChild`s or required `Input`s etc). In the other places it would probably require more refactoring to get rid of it but at least here it would be easy to convert it to `loading = false`

aorin (changes requested):
Good job for continuing this! Since only the part of the project is refactored `tsconfig.app.json` and `tsconfig.json` files can't be in this PR as the app wouldn't compile anymore?

projects/laji/src/app/shared/navbar/notifications/notifications.component.ts
olzraiti: I set it as `loading = true`, which makes more sense. It's loading from the beginning, until it's loaded. UI wise this has no difference, seems like the whole component isn't displayed before the facade's loading observable sends a value.

olzraiti:
Good catch, I reverted the ts config file changes.

aorin (approved):
:+1:

PR #483, opened 2024-01-17
Fix project form lang change [#186797710]

olzraiti:
https://www.pivotaltracker.com/story/show/186797710
https://186797710.dev.laji.fi

# Why?

Fixes this bug:

1. navigate to trip report form
2. change lang
-> document form lang doesn't change

# How ?

* rm lang param from form service in order to simplify things (it's always just the current lang)
* make form service cache to take lang into account
* project form service flushes "current form" on lang change


projects/laji/src/app/shared/service/form.service.ts
```diff
@@ -37,48 +39,50 @@ export class FormService {
...
-  getFormInJSONFormat(formId: string, lang: string): Observable<Form.JsonForm> {
+  getFormInJSONFormat(formId: string): Observable<Form.JsonForm> {
     if (!formId) {
       return ObservableOf({} as Form.JsonForm);
     }
-    this.setLang(lang);
-    if (!this.jsonFormCache[formId]) {
-      this.jsonFormCache[formId] = this.lajiApi.get(LajiApi.Endpoints.forms, formId, {lang, format: 'json'}).pipe(
+    const lang = this.translate.currentLang;
+    const cacheKey = getCacheKey(formId, lang, 'json');
```
aorin, line 70: The format parameter seems redundant since it already uses different caches for different forms (`formCache` and `jsonFormCache`)

projects/laji/src/app/shared/service/form.service.ts
```diff
@@ -37,48 +39,50 @@ export class FormService {
     private http: HttpClient,
     private userService: UserService,
     private translate: TranslateService
-  ) {}
+  ) {
+    this.translate.onLangChange.subscribe(this.setLang);
```
aorin, line 43: The return type of this is not the language so it should be more like `(event) => this.setLang(event.lang)`

aorin (changes requested):

aorin (approved):
:+1:

PR #484, opened 2024-01-17
Fix observation form sidebar sticky header

Blodir:
Fixes the issue with observation form sidebar header (search button section) that was mentioned today, i.e. it now stays on top correctly when the sidebar is scrolled down.

Related: https://github.com/luomus/laji/pull/472

`display: flex` is set in the default styles in scss so now angular animations matches it...

Ideally the open state in angular animations would just unset the display property, however attempting to do that inexplicably breaks SSR hydration (the sidebar will stay at `display: none`).


olzraiti (approved):
LGTM

PR #486, opened 2024-01-17
Make filters only use titles in selected language

Blodir:
https://www.pivotaltracker.com/story/show/186331507

With some added type magic on the side...

### DIFF
--- a/projects/laji/src/app/shared/service/filter.service.ts
+++ b/projects/laji/src/app/shared/service/filter.service.ts
@@ -1,10 +1,12 @@
 import { Injectable } from '@angular/core';
 
-type FilterBaseType = number|boolean|string;
-
-interface FilterObjType {by: FilterBaseType; properties: string[] }
-
-export type FilterByType = FilterBaseType|FilterObjType;
+export type FilterBaseType = number | boolean | string;
+export interface SearchRecordQuery<T, K> { by: T; properties: K };
+export type FilterByType<
+  T extends FilterBaseType = FilterBaseType,
+  K extends string[] = string[]
+> = T | SearchRecordQuery<T, K>;
+export type SearchRecord<K extends string[]> = { [P in K[number]] } & Record<string, any>;
 
 @Injectable({
   providedIn: 'root'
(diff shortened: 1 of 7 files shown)


projects/laji/src/app/shared/service/filter.service.ts
```diff
@@ -13,26 +15,44 @@ export class FilterService {
...
+    Element extends T | S
+  >(
+    arr: Element[], filterBy: FilterByType<T, K>, matching = true
+  ): Element[] {
+    if (!Array.isArray(arr)) {
+      return [arr];
     }
-    let needle: any = filterBy;
-    let properties: any = null;
+    if (!filterBy && typeof filterBy !== 'boolean') {
```
olzraiti, line 33: `filterBy === false` instead of  `typeof filterBy !== 'boolean'` would be more readable at least for me.

olzraiti (changes requested):
The typings look good :+1: left just a minor nit.

I also noticed a bug which I'm not sure if it could cause problems in the real world (I guess the filter is ran for some JSON values so it could be a problem because JSON can be serialized into `null`):

```
this.filter([{a: 1}, {a: null}], false);
```

This will run into a null pointer error because `typeof null === 'object'). It was broken already, so no need to fix here I guess.

projects/laji/src/app/shared/service/filter.service.ts
Blodir: These are not equivalent. The if condition says that filterBy must be falsy, but not a boolean. The suggested change says that filterBy must be falsy and must be false (i.e. it has to be boolean).

Blodir: The aforementioned `this.filter([{a: 1}, {a: null}], false);` input should not pass type check, but it does, which is a bug... I think maybe it would be the best to move on for now though instead of spending more time strengthening the typing here :D

olzraiti: Right, yeah I meant `filterBy !== false`. But idk, that's not objectively better. Maybe this wasn't the most productive review, let's just get on with it... :D

olzraiti (approved):

PR #495, opened 2024-01-19
Taxonomy sidebar cannot be closed on mobile 186579345

MLumme:
Task here https://www.pivotaltracker.com/story/show/186579345, testable here https://186579345.dev.laji.fi/

### DIFF
--- a/projects/laji-ui/src/lib/sidebar/sidebar.component.html
+++ b/projects/laji-ui/src/lib/sidebar/sidebar.component.html
@@ -53,7 +53,7 @@
 </ng-template>
 
 <ng-template #closeBtn>
-  <div class="close-wrapper">
+  <div class="close-wrapper" [ngClass]="{'mobile': mobile}">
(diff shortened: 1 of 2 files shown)


projects/laji-ui/src/lib/sidebar/sidebar.component.html
```diff
@@ -53,7 +53,7 @@
 </ng-template>
 
 <ng-template #closeBtn>
-  <div class="close-wrapper">
+  <div class="close-wrapper" [ngClass]="{'mobile': mobile}">
```
aorin, line 56: Seems unnecessary to add the class here since the sidebar wrapper already has that class. The css selector can be `.mobile .close-wrapper`

aorin (changes requested):

projects/laji-ui/src/lib/sidebar/sidebar.component.html
MLumme: It actually does need it to trigger the subclass of close-wrapper which overrides the button position in close-wrapper itself.

aorin (approved):
Thanks!

PR #499, opened 2024-01-22
Playwright +observation and +save-observations migration

rpulkka:
Playwright migrations for +observation and +save-observations tests. All tests should use auto-retrying assertions. Most should not be flaky.


projects/laji/src/app/shared-modules/laji-map/laji-map.component.ts
```diff
@@ -166,6 +119,32 @@ export class LajiMapComponent implements OnDestroy, OnChanges {
...
+      }
+
+      if (changes.options) {
+        this.updateOptions();
+      }
+
+      let settings$ = of(this.userSettings);
+
+      if (changes.settingsKey) {
+        const key = this.settingsKey;
```
aorin, line 135: This could still be little refactored. Looks like `this.settingsKey` and `this._settingsKey` has always the same value so `_settingsKey` can be removed and also the `key` variable can be removed as it's just the `settingsKey`

projects/laji/test/+observation/observation.po.ts
```diff
@@ -0,0 +1,208 @@
...
+
+  async navigateTo(sub: 'list' | '' = '', query?: Record<string, string>) {
+    await this.page.goto(`observation/${sub}?${new URLSearchParams(query || {}).toString()}`);
+  }
+
+  async navigateToViewWithAllFilters() {
+    await this.page.goto(this.pathWithAllFilters);
+  }
+
+  async hasGraphqlErrors() {
```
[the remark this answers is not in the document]
aorin: Now it should work but we had discussion on the other PR (https://github.com/luomus/laji/pull/481) that had similar thing that the `hasGraphQLApiErrors` variable can be a little confusing, maybe this could also be converted to a function like here for consistency https://github.com/luomus/laji/pull/481/commits/5351376f87eda7fe9cb8d1df24ddd6dd08d67b33

projects/laji/test/+save-observations/save-observations.po.ts
```diff
@@ -3,6 +3,7 @@ import { NavPage } from '../shared/nav.po';
 
 export class SaveObservationsPage {
 
+  _page = this.page;
```
aorin, line 6: No need to have an extra variable, just set the constructor variable public (`constructor(public page: Page) { }`)

aorin (changes requested):

aorin (approved):
Thanks, looks good

PR #500, opened 2024-02-05
Remove webworker login

Blodir:
https://www.pivotaltracker.com/story/show/186962721
https://8972345.dev.laji.fi/

### DIFF
--- a/projects/laji/src/app/shared/service/user.service.ts
+++ b/projects/laji/src/app/shared/service/user.service.ts
@@ -50,7 +50,7 @@ export interface UserServiceState extends PersistentState {
   allUsers: {[id: string]: Person|Observable<Person>};
 }
 
-export const createProfile = (profile: Profile | null, user: Person | null): Profile => {
+export const prepareProfile = (profile: Profile | null, user: Person | null): Profile => {
   if (!profile) {
     profile = {};
   }
@@ -261,7 +261,7 @@ export class UserService implements OnDestroy {
       this.personApi.personFindProfileByToken(this.getToken()),
       this.user$
     ]).pipe(
-      map(([profile, user]) => createProfile(profile, user)),
+      map(([profile, user]) => prepareProfile(profile, user)),
       take(1)
     );
   }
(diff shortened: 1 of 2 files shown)


projects/laji/src/app/shared/service/user.service.ts
```diff
@@ -25,96 +25,113 @@ import { PlatformService } from '../../root/platform.service';
...
+export interface UserServiceState extends PersistentState {
   user: Person | null;
-  settings: IUserSettings;
+  settings: UserSettings;
   allUsers: {[id: string]: Person|Observable<Person>};
 }
 
-const _persistentState: IPersistentState = {
-  isLoggedIn: false
+export const createProfile = (profile: Profile | null, user: Person | null): Profile => {
```
olzraiti, line 57: I think `prepareProfile` was a better name. I associate the word "create" with actually creating some external entity, whereas "prepare" I associate with modifying some entity for internal usage. Could also be just simply "profile()", since it's about "serializing" it into our internal model of the profile.

projects/laji/src/app/shared/guards/check-login.guard.ts
```diff
@@ -23,38 +22,24 @@ export class CheckLoginGuard  {
...
     }
 
-    return this.userService.checkLogin();
+    this.isLoginChecked = true;
+
+    // store in localstorage
+    return this.userService.login(route.queryParams['token']).pipe(map(_ => true));
+
+    // TODO go to return url
+    return this.router.parseUrl(this.userService.getReturnUrl());
```
olzraiti, line 43: The TODO here must be implemented, it's an important feature.

olzraiti (commented):
GJ on refactoring some interface names etc. I left one complaint about that.

I'm a bit undecided about stuff removed from being static methods of classes - it has the benefit of making functions more discoverable since you can find them by autocomplete, and helps with name spacing. Either way is fine by me, as long as it's consistent. We should maybe have some mutually agreed code style guide for things like this.

Login not redirecting back to the page where login happened must be implemented.

projects/laji/src/app/shared/guards/check-login.guard.ts
Blodir: Not sure how I missed this... should work now

olzraiti:
Thanks!

olzraiti (approved):

PR #504, opened 2024-03-04
Add clarifying text for docs where sensitive observations were removed

Blodir:
https://www.pivotaltracker.com/story/show/186844044
https://186844044.dev.laji.fi/


MLumme (approved):
LGTM, with the reservation that nothing to test it with  is easily found.

PR #506, opened 2024-03-04
Fix horizontal scroll of col headers and rows being unsynced

Blodir:
https://www.pivotaltracker.com/story/show/185990790
https://185990790.dev.laji.fi/


MLumme (approved):
LGTM

PR #507, opened 2024-03-04
Get informal group imgs from cdn

Blodir:
https://www.pivotaltracker.com/story/show/186579256
https://186579256.dev.laji.fi/taxon/browse


rpulkka (approved):
Tested on Safari/Macbook.

PR #508, opened 2024-03-04
Column selection modal improvements

Blodir:
https://www.pivotaltracker.com/story/show/186579316
https://186579316.dev.laji.fi/observation/list


rpulkka (changes requested):
I would prefer the numbers to be centered between the arrow selection instead of leaning left (flex-start). Would this be easy to fix?

Blodir:
yep, trivial change 👍

projects/laji/src/app/shared-modules/observation-result/observation-table/observation-table-settings.component.ts
```diff
@@ -9,9 +9,7 @@ import {ModalRef, ModalService} from 'projects/laji-ui/src/lib/modal/modal.servi
   selector: 'laji-observation-table-settings',
   template: `
```
rpulkka, line 10: Not a fan of including the template in the ts-file, but probably it isn't a problem, and it was there before this PR anyway.

projects/laji/src/app/shared-modules/observation-result/selected-field-item/selected-field-item.component.ts
```diff
@@ -18,6 +18,10 @@ export class SelectedFieldItemComponent {
   @Output() moveUp = new EventEmitter<string[]>();
   @Output() moveDown = new EventEmitter<string[]>();
 
+  ngOnInit() {
+    console.log(this.field, this.label, this.idx, this.len, this.required);
```
rpulkka, line 22: console.log should probably be removed?

rpulkka (changes requested):
There's a console.log, not sure if it was left by accident or purpose. Also, maybe not in the scope of this PR, but the arrow icon inside the left and right buttons looks to lean to the bottom instead of center, but this issue was present before this PR also. So either fix or write a Pivotal task for later?

projects/laji/src/app/shared-modules/observation-result/observation-table/observation-table-settings.component.ts
Blodir: yep i agree, i'll move it anyways

Blodir:
I decided to vertically align the arrow even though there's really no 'pretty' way of doing it because it's a font symbol arrow

rpulkka (approved):
Now it looks good to me (both code and UI)!

PR #512, opened 2024-03-13
Bsg changes

aorin:
Can be tested at: https://kerttu-global-dev.laji.fi/ or https://bsg.laji.fi/

Some changes to the BSG project, shouldn't affect other projects.


rpulkka (approved):
Looks good to me!

PR #513, opened 2024-04-05
Named places 'all' filter as empty to API query

olzraiti:
Currently the named places filtering works kind of accidentally. Let's look at the invasive species named places view for example:

https://dev.laji.fi/en/project/MHL.33/form/places

If you select some municipality, and then again 'All municipalities', the API query will have `&municipality=all`. The old API has some logic that filters non QNames for some query params. It's implemented in a very messy way, and I also just don't agree that we should do that in the API. So, in the new API I'm not going to do that kind of filtering for the params - if anything, I'd rather throw an exception for a bad query param. But I don't think that's really necessary. So, this PR fixes the query so that it won't have `municipality=all`. Instead, there just won't be a municipality filter (same as all).

Testable at https://240405.dev.laji.fi/en/project/MHL.33/form/places

Includes also:
    * `id` isn't anymore automatically added to the `selectedFields` of the named place query by the API, so I added it here.
    * null pointer error in named place view, when user is undefined (broken atm eg https://beta.laji.fi/project/MHL.1/form/MHL.1/places when unlogged)

### DIFF
--- a/projects/laji/src/app/+project-form/form/named-place/named-place/named-place.component.html
+++ b/projects/laji/src/app/+project-form/form/named-place/named-place/named-place.component.html
@@ -71,7 +71,7 @@
           <laji-np-choose [namedPlaces]="data.namedPlaces"
                           [documentForm]="data.documentForm"
                           [placeForm]="data.placeForm"
-                          [userID]="data.user.id"
+                          [userID]="data.user?.id"
                           [activeNP]="data.activeNP?.id"
                           [formRights]="data.formRights"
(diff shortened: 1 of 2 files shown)


olzraiti:
Personally I'd like it to not to be there. I recall that I tried removing but it seemed to require more extensive refactoring, so I left it out as scope creeping - this PR is to address a change in the API

MLumme:
Aight, approved.

MLumme (approved):

PR #514, opened 2024-04-12
Playwright

aorin:
Migrates Protractor tests to Playwright


Blodir (approved):
I only looked at the changes to login, since I think that is the only thing that is not reviewed in this PR (?), but looks good!

PR #516, opened 2024-04-23
Modify-button visible for secondary data #186261663

MLumme:
Task: https://www.pivotaltracker.com/story/show/186261663, testable here: https://186261663.dev.laji.fi/.

Solving this required adding a query to form endpoint to read the field denoting a form as member of secondary collection, which I'm not too keen on.


olzraiti (changes requested):
Could be slightly optimized so that it doesn't fetch a single forms, but instead checks the option from the form listing. It's a good way to check options, because the query has to be done only once for multiple different document viewers / any other component that needs to know about the form options. The method is `formService.getFormInListFormat`

Rm the console.log

MLumme:
> getFormInListFormat

Done, much better method.

olzraiti (approved):
:+1:

PR #520, opened 2024-05-29
Fix document viewer checking write rights [#187698581]

olzraiti:
The document viewer dispalyed the document writable (= shows edit & delete buttons) based on just looking at the form's secondaryCopy option. This made all documents with non-secondaryCopy form display the buttons. I made it to act on the secondaryCopy option only if it's false.

Also removed few unused variables from the document annotation component.

### DIFF
--- a/projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
+++ b/projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
@@ -183,10 +183,10 @@ export class DocumentComponent implements AfterViewInit, OnChanges, OnInit, OnDe
       switchMap(doc => doc.doc.formId
         ? this.formService.getFormInListFormat(IdService.getId(doc.doc.formId)).pipe(
           map(form => {
-            const isSecondary = !!form.options?.secondaryCopy;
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
```diff
@@ -198,6 +198,7 @@ export class DocumentComponent implements AfterViewInit, OnChanges, OnInit, OnDe
       .subscribe(({doc, rights}) => {
           this.hasEditRights = rights.hasEditRights;
           this.hasDeleteRights = rights.hasDeleteRights;
+        console.log(rights);
```
aorin, line 201: This should be removed

aorin (changes requested):

projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
olzraiti: Thanks, removed also another console.log elsewhere in the same commit

aorin (approved):

PR #523, opened 2024-06-03
Bird count result service improvements [#187504768]

rpulkka:
https://187504768.dev.laji.fi/project/MHL.65/stats
https://www.pivotaltracker.com/n/projects/2346653/stories/187504768

Implements the bird count result service improvements: Also 0 count squares are shown on map overlay. Selected taxon persists when changing between stats and map on the bird point count result service. Polygons with more observations are shown on top of polygons with less observations so that actual points of interest are not being hidden by empty squares' sides/corners.

Failed to implement query for multiple taxonSets from GET /taxon endpoint because the taxonSets -param is broken and we need to wait Esko to fix the Java Laji-Backend for that because we can't set up the environment.


projects/laji/src/app/+project-form/results/common/result-map/result-map.component.ts
```diff
@@ -112,6 +112,62 @@ export class ResultMapComponent implements OnInit {
 
   toHtmlSelectElement = toHtmlSelectElement;
 
+  private gatheringCountMapData$: Observable<DataOptions> = of([this.getGatheringCounts$(), this.getGatheringCounts$(true)]).pipe(
```
Blodir, line 115: Can you move long definitions like this into the constructor? So above the constructor just declare `private gatheringCountMapData$: Observable<DataOptions>;` and move the definition to `this.gatheringCountMapData$ = ...` in the constructor

Blodir (changes requested):

projects/laji/src/app/+project-form/results/common/result-map/result-map.component.ts
```diff
@@ -182,6 +129,62 @@ export class ResultMapComponent implements OnInit {
     this.mapOptions = {
       clickBeforeZoomAndPan: true
     };
+
+    this.gatheringCountMapData$ = of([this.getGatheringCounts$(), this.getGatheringCounts$(true)]).pipe(
```
Blodir, line 133: The value inside of `of()` is just being thrown away, so you might as well call of with no arguments.

Blodir (changes requested):

projects/laji/src/app/+project-form/results/common/result-map/result-map.component.ts
Blodir: Goes for the other one too

PR #526, opened 2024-06-27
Use laji-ui placement service in typeahead instead of ngx-bootstrap

Blodir:
Basically circumvents the positioning service that was copy pasted from ngx-bootstrap entirely. Instead using the placement service that I wrote some time ago for tooltips/popovers/etc.

The typeahead is an absolute monster. It's difficult to comprehend how it's possible to have thousands of LOC for what amounts to a simple dropdown. It seems really bad performance wise as well. Eg. it doesn't use `OnPush` and I checked that it was calling the positioning logic constantly even when no change was happening...

I added some checks for invalid values, but in practice I never saw them come up when testing. The z-index is to ensure that the dropdown opens on top always (such as on new annotation form).

https://www.pivotaltracker.com/n/projects/2346653/stories/187852117
https://187852117.dev.laji.fi/

List of affected components (stuff that uses the typeahead):
- observation form: taxon
- observation form: observer name
- new annotation form
- "find person"
- taxon autocomplete
  - excel import mapping
  - line transect result
- taxon select (why do we have two?)
  - species form (the filters on the list)
  - species info card
  - invasive species control project
  - iucn filters (?)
  - iucn about page
- kerttu species select

### DIFF
--- a/projects/laji-ui/src/lib/typeahead/positioning/positioning.service.ts
+++ b/projects/laji-ui/src/lib/typeahead/positioning/positioning.service.ts
@@ -46,7 +46,7 @@ export interface PositioningOptions {
 export class PositioningService {
   private options?: Options;
   private update$$ = new Subject<null>();
-  private positionElements = new Map();
+  private positionElements = new Map<HTMLElement, PositioningOptions>();
   private triggerEvent$?: Observable<number|Event|null>;
   private isDisabled = false;
 
@@ -72,21 +72,9 @@ export class PositioningService {
             return;
           }
 
-          // We could call placementService.update here, but I don't think it's necessary
-          // The old ngx-bootstrap positioning logic has been entirely replaced
-          return;
-          this.positionElements
-          // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            .forEach((positionElement: any) => {
-              positionElements(
-                _getHtmlElement(positionElement.target),
-                _getHtmlElement(positionElement.element),
-                positionElement.attachment,
-                positionElement.appendToBody,
-                this.options,
-                rendererFactory.createRenderer(null, null)
-              );
-            });
+          for (let el of this.positionElements.keys()) {
+            this.placementService.update(el);
+          }
         });
       });
     }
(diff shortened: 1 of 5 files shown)


projects/laji-ui/src/lib/typeahead/positioning/positioning.service.ts
```diff
@@ -69,6 +72,9 @@ export class PositioningService {
             return;
           }
 
+          // We could call placementService.update here, but I don't think it's necessary
+          // The old ngx-bootstrap positioning logic has been entirely replaced
+          return;
           this.positionElements
```
aorin, line 78: I think it would be clearer to comment out the whole section that is not used anymore or remove it altogether as it can always be found in the history if it's needed. I'm also wondering if this really is unnecessary code, I tried to resize the window while the typeahead was open and it didn't change place, should it?

projects/laji-ui/src/lib/typeahead/positioning/positioning.service.ts
```diff
@@ -86,8 +92,8 @@ export class PositioningService {
     }
   }
 
-  position(options: PositioningOptions): void {
-    this.addPositionElement(options);
+  position(options: PositioningOptions, renderer: Renderer2): void {
```
aorin, line 95: Is there a need to pass the renderer as a parameter or could it use the `rendererFactory` like before. Like this for example:
```
private renderer: Renderer2;
constructor(...) {
    ...
    this.renderer = rendererFactory.createRenderer(null, null);
}
```

aorin (changes requested):
Seems

projects/laji-ui/src/lib/typeahead/positioning/positioning.service.ts
Blodir: I know passing the renderer as a parameter makes the code a bit more cumbersome to write, but I don't think new renderers should be created unless absolutely necessary (which it isn't in this case).

aorin (approved):
:+1:

PR #528, opened 2024-07-03
improve about pages on mobile

Blodir:
https://www.pivotaltracker.com/story/show/187765471
https://187765471.dev.laji.fi/about/5719


rpulkka (approved):
Good! Also tested that scrolling works on phone.

PR #529, opened 2024-07-03
Change protax api url

aorin:
Test that the new url works: go to https://32.dev.laji.fi/theme/protax, switch to the "File selector" tab and upload the [protax-example-10.txt](https://github.com/user-attachments/files/16086135/protax-example-10.txt) file. The new urls are connected to Rahti 2.


rpulkka (approved):
OK!

PR #530, opened 2024-07-18
Bird count result service improvements pt. 2 [#187725477]

rpulkka:
https://187725477.dev.laji.fi/project/MHL.65/stats
https://www.pivotaltracker.com/n/projects/2346653/stories/187725477

Added a dropdown for selecting year on the result service map (bird result services start from 2019 and biomon from 2023). "All years" shows observations from starting year to now. Also fixed the loading spinner behavior.


projects/laji/src/app/+project-form/results/biomon-result/biomon-result.component.ts
```diff
@@ -116,4 +117,8 @@ export class BiomonResultComponent implements OnInit, OnDestroy {
   onTaxonChange(taxon: any) {
     this.updateState({ taxon });
   }
+
+  onYearChange(year: any) {
```
Blodir, line 121: could this be typed?

projects/laji/src/app/shared/api/WarehouseApi.ts
```diff
@@ -195,6 +196,13 @@ export class WarehouseApi {
     return this.warehouseQueryGet(target, query, aggregateBy, orderBy, pageSize, page, geoJSON, onlyCount);
   }
 
+  // query to unit/aggregate endpoint with front-end cache
+  @cacheReturnObservable(120000)
+  public warehouseQueryAggregateGetCached(query: WarehouseQueryInterface = {}, aggregateBy?: Array<string>, orderBy?: Array<string>, pageSize?: number, page?: number, geoJSON?: boolean, onlyCount?: boolean): Observable<PagedResult<any>|any> {
```
Blodir, line 201: I think you should cache `getGatheringCounts$` directly instead. Different uses of `warehouseQueryAggregateGet` may require different cache times

Blodir (changes requested):

projects/laji/src/app/+project-form/results/common/result-map/result-map.component.ts
```diff
@@ -120,75 +125,87 @@ export class ResultMapComponent implements OnInit {
...
-                +item.lat,
-                +item.lon,
-                { count: item.count }
-              )
-            );
-            return _features;
-          }, []).sort((a, b) => a.properties.count - b.properties.count)
-        }
-      })),
+      switchMap(([collections, taxon, year]) => this.getGatheringCounts$(collections, taxon, year).pipe(
```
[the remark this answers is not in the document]
rpulkka: Tried to flatten the structure not sure if it's correct now.

projects/laji/src/app/shared/api/WarehouseApi.ts
rpulkka: Reason why I failed at first was the all taxaParam which caused same query to be called with different params -> no cache hit. Also taxon and year could be either string of undefined, so I tried to address this issue in the latest commit.

Blodir:
Looks nice, good job!

Blodir (approved):

PR #532, opened 2024-08-19
User service fixes

Blodir:
- Fixes blank page when page should be redirected by the `OnlyLoggedIn` guard
- Fixes logout always erroring, because it was expecting json from the api (though the logout did succeed)
- Refactored user service to use tagged unions to represent loading states


olzraiti (approved):
LGTM, GJ on writing more robust state management.

PR #533, opened 2024-08-20
Fix days missing from the day chart

aorin:
https://www.pivotaltracker.com/story/show/188126950
https://188126950.dev.laji.fi/observation/statistics


rpulkka (approved):
OK!

PR #535, opened 2024-08-26
Validate user service local storage state

Blodir:
Fixes https://luomus-ict.slack.com/archives/CQEQSQCP7/p1724406648826119

Also gave the ready user state a more descriptive tag

https://762038495.dev.laji.fi/

### DIFF
--- a/projects/laji/src/app/shared/service/user.service.ts
+++ b/projects/laji/src/app/shared/service/user.service.ts
@@ -128,11 +128,11 @@ export const getLoginUrl = (next = '', lang = DEFAULT_LANG, base = '') => {
 
 export const isIctAdmin = (person: Person): boolean => person?.role?.includes('MA.admin') ?? false;
 
-const validateKeysRecursively = (validatee: Object, validator: Object): boolean => {
+const validateKeysRecursively = (validatee: any, validator: any): boolean => {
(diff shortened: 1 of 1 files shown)


olzraiti (changes requested):
The logic is difficult to follow because it tries to prevent problems created by the old format of the local storage's `userState`. I think taking the old format to account in the typings would help. The `userState` variable will remain in the old format for a long time for some users, as some users might log in after years. We should type our app against what the "rest of the world" actually is instead of what we wish it was.

Suggestion to ensure type safety:

1. Type `localStoragePersistentState` as `PersistentState | LegacyPersistentState | undefined` or even just `unknown`. If it's unknown, maybe add a comment that the format is changed and we can't know it's type before validating it.
2. Make `get persistentState` always return a `PersistentState`. Do the validation `validateKeysRecursively` here. This way the `persistentState` is always safe to use.
3. `validateKeysRecursively` is unnecessarily complex for the use case. Replace it with a `isPersistentState(state: PersistentState | LegacyPersistentState | undefined): is PersistentState` or `isPersistentState(state: unknown): is PersistentState`. This would automatically help with narrowing the type of the validated object. The logic can also be simplified a lot - maybe just `const isPersistentState = (state: unknown): is PersistentState => state?.loginState`

PR #536, opened 2024-08-28
user$ should return undefined when user is not logged in

Blodir:
Two issues:
- user service `user$` behavior didn't exactly correspond to old one after refactor. It used to return `Person | undefined` and now it was returning just `Person`
- there was a bug in that the person part was in loading state instead of logged out state

https://42398123.dev.laji.fi/


olzraiti (approved):
LGTM

PR #538, opened 2024-09-02
Autocomplete opens on top of input

Blodir:
https://www.pivotaltracker.com/story/show/188206822
https://188206822.dev.laji.fi/observation


rpulkka (approved):
OK!

PR #539, opened 2024-09-06
Bird atlas update

Blodir:
Contains various updates to the bird-atlas result service frontend which are already in production.


rpulkka (approved):
OK!

PR #542, opened 2024-09-10
Shp deprecation notice

aorin:
https://www.pivotaltracker.com/story/show/188161910

Adds deprecation notice of the shp format to citable downloads (https://188161910.dev.laji.fi/citation/HBF.11712) and light downloads (https://188161910.dev.laji.fi/observation/list?target=MX.48089)


MLumme (approved):
LGTM

PR #548, opened 2024-09-25
fix polygon registration catch block path to error [closes #547]

### LINKED ISSUE #547: Observation map cannot search after drawing a feature
### Repro

1. Goto https://laji.fi/observation/map
2. Draw a new polygon ("piirrä kuvio kartalle")

### What happens?

Nothing!

### What should happen?

The "Hae" button should be enabled

olzraiti:
Fixes a bug where you draw a large polygon to the observation map and nothing seems to happen.

Not sure if the catch block has ever worked. I tested the behaviour against the old API, and it doesn't work even there so idk...


olzraiti:
Closes #547

Blodir (approved):

PR #549, opened 2024-09-26
Cannot add two friends #184178674

MLumme:
https://www.pivotaltracker.com/story/show/184178674
https://184178674.dev.laji.fi/

Added change detection to clear out the variable locking the button to add a new friend when page changes to a new user, unfortunately now if user uses back button they can resend the request, but that was possible already before via reloading the page.


rpulkka (approved):
OK!

PR #553, opened 2024-09-26
Sidebar displays observations from another form #187298216

MLumme:
https://www.pivotaltracker.com/story/show/187298216
https://187298216.dev.laji.fi/vihko/home

Fixed by hiding the latest documents in sidebar until the latestDocumentFacade is done loading if the call to it is first since page load, to keep documents still in latestDocumentFacade potentially from other projects from showing up. LatestDocumentFacade also could cause page to update with wrong projects documents after pagechange due to hanging subscription to latest documents, so I added unsuscribe call to it when the query changes.


rpulkka (approved):
OK!

PR #555, opened 2024-09-26
Accuracy filter slider at incorrect position #184453892

MLumme:
https://www.pivotaltracker.com/story/show/184453892
https://184453892.dev.laji.fi/observation/list?coordinateAccuracyMax=100

Added initalization from query parameter to the accuracy slider, before it was always at default value regardless of query parameters value, removed box shadow from the accuracy and coordinate overlap slider due to chrome bug, added margin-to to coordinate-sections hr-separators to more clearly separate different subsections.


projects/laji/src/app/+observation/form/observation-form.component.scss
```diff
@@ -25,13 +25,13 @@ laji-select, laji-metadata-select {
 hr {
   margin: 4px 0;
 }
-#logCoordinateAccuracyMax:focus {
+#coordSeparator {
```
Blodir, line 28: ids are exclusive (you shouldn't have two elements with the same id on the same page). Use a class instead

projects/laji/src/app/+observation/form/observation-form.component.html
```diff
@@ -329,7 +329,7 @@
                 max="5"
                 list="accuracyPowers"
                 (click)="enableAccuracySlider()"
-                (change)="onAccuracySliderChange()"
+                (input)="onAccuracySliderChange()"
```
Blodir, line 332: I'd probably change the name here to reflect that the event has changed to `input` i.e. `onAccuracySliderInput()`

Blodir (changes requested):

MLumme:
@Blodir Implemented your changes, and  moved the codeblock that sets the accuracy value from parameters to a different function, the current form stopped working, and seemed to have worked only by accident.

Blodir (approved):

PR #573, opened 2024-10-04
Fix null pointer error for undefined user in np-info

olzraiti:
Repro:

1. goto http://beta.laji.fi/project/MHL.1/form/MHL.1/places?activeNP=MNP.27447 when unlogged
-> Error is thrown

Fix testable in https://241004.dev.laji.fi/project/MHL.1/form/MHL.1/places?activeNP=MNP.27447

There's something wrong in the user service logic. Even though the `user$` is stated as `Observable<Person | undefined>`, if you look at the actual type of it (hover in you IDE or however you do it), it's just `Observable<Person>`. This is interesting, as I though stating a type has stronger preference than type inference.

There might be similar errors somewhere as the typing is wrong (`user$` can emit `undefined` even though the typing says otherwise), but I don't have time to dig deeper just now.

### DIFF
--- a/projects/laji/src/app/+project-form/form/named-place/np-info/np-info.component.ts
+++ b/projects/laji/src/app/+project-form/form/named-place/np-info/np-info.component.ts
@@ -213,6 +213,9 @@ export class NpInfoComponent implements OnInit, OnChanges, AfterViewInit {
       return;
     }
     this.userService.user$.pipe(take(1)).subscribe(person => {
+      if (!person) {
+        return;
+      }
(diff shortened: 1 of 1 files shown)


Blodir (approved):

Blodir:
Yeah ts_ls gives the type `Observable<Person>` for me too... Not sure what causes this, but I think there's a big chance that it's because we allow implicit undefined and that it would be fixed in strict mode

olzraiti:
@Blodir yeah that might be the case indeed :+1:

Issue #571, opened 2024-10-04
Laji.fi layout issues


mikkohei13:
1) Species page overview tab map and GBIF map are not centered and partly grey. Firefox / Win 10.

![Image](https://github.com/user-attachments/assets/de070f80-60a5-490e-9aea-7b4c32d4b146)

2) Species page biology tab header ovelapping. Firefox / Android

![Image](https://github.com/user-attachments/assets/bf2275e2-c60e-4c09-b8b9-b425117ca18d)

3) Species page taxonomy tab few headings are off. Firefox/Win 

![Image](https://github.com/user-attachments/assets/4eba30c5-833b-40e9-b0b9-f0a1c6675f78)

PR #576, opened 2024-10-08
Fix various bugs with the common panel/accordion component

Blodir:
Related (copied from slack):

Ramble about angular animations: There have been various issues with the panels including this one, things overlapping on the observation page, taxon info card maps breaking, etc. These are all caused by angular animations (especially in combination with ssr/hydration).

Issues:
1. the SSR generated html includes invalid syntax like style=""height=*;" (this comes from angular animations wildcard).
2. After hydration completes animation state doesn't update even if the predicate changes during hydration
3. Since the animation targets the height property, sometimes the height updates correctly, but no resize event is emitted to child nodes (which breaks the map)
4. Other stuff I guess...

We do have animations on some other components that seem to not have issues (eg. sidebar), but I can't seem to figure out why it breaks for laji-panel . Luckily the animations are pretty unnecessary for the component so I will just remove them. These issues have been horrible to debug.

https://www.pivotaltracker.com/story/show/188283099
Fixes bullet points 1 and 3.

### DIFF
--- a/projects/laji/src/app/shared/panel/panel.component.html
+++ b/projects/laji/src/app/shared/panel/panel.component.html
@@ -6,11 +6,7 @@
     <strong class="panel-title" [innerHtml]="title"></strong>
   </ng-template>
 </div>
-<div class="panel-collapse" [ngClass]="{closed: hideInside}"
-     (@visibilityState.done)="animationDone($event)"
-     (@visibilityState.start)="animationStart($event)"
-     [@visibilityState]="open ? 'in' : 'out'"
-     role="tabpanel" aria-labelledby="">
+<div class="panel-collapse" [ngStyle]="{display: open ? 'block' : 'none' }" role="tabpanel" aria-labelledby="">
   <div class="panel-body">
     <ng-content></ng-content>
   </div>
--- a/projects/laji/src/app/shared/panel/panel.component.html
+++ b/projects/laji/src/app/shared/panel/panel.component.html
@@ -6,7 +6,7 @@
     <strong class="panel-title" [innerHtml]="title"></strong>
   </ng-template>
 </div>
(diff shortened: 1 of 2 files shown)


projects/laji/src/app/shared/panel/panel.component.html
```diff
@@ -6,11 +6,7 @@
     <strong class="panel-title" [innerHtml]="title"></strong>
   </ng-template>
 </div>
-<div class="panel-collapse" [ngClass]="{closed: hideInside}"
-     (@visibilityState.done)="animationDone($event)"
-     (@visibilityState.start)="animationStart($event)"
-     [@visibilityState]="open ? 'in' : 'out'"
-     role="tabpanel" aria-labelledby="">
+<div class="panel-collapse" [ngStyle]="{display: open ? 'block' : 'none' }" role="tabpanel" aria-labelledby="">
```
olzraiti, line 9: `aria-labelledby=""` is unnecessary

projects/laji/src/app/shared/panel/panel.component.ts
```diff
@@ -1,48 +1,19 @@
 /* eslint-disable @angular-eslint/component-selector */
-import { Inject, PLATFORM_ID, AfterViewInit, ChangeDetectorRef, Component, EventEmitter, Input, NgZone, OnChanges, Output, TemplateRef } from '@angular/core';
-import { animate, state, style, transition, trigger } from '@angular/animations';
-import { isPlatformBrowser } from '@angular/common';
+import {Component, EventEmitter, Input, Output, TemplateRef } from '@angular/core';
```
olzraiti, line 2: missing space after `{`

olzraiti (changes requested):

olzraiti (approved):

Blodir:
Fixes #571

PR #606, opened 2024-10-22
Strict TS config - Week 1

rpulkka:
Could someone review my first week's strict TS config changes? Not sure what to do in situations where a type is now possibly undefined, and it isn't addressed in the code, (see newline 220 on own-datatable.component.ts, or code in technical-news-dumb.component.ts).

### DIFF
--- a/projects/laji/src/app/shared-modules/own-submissions/own-datatable/own-datatable.component.ts
+++ b/projects/laji/src/app/shared-modules/own-submissions/own-datatable/own-datatable.component.ts
@@ -146,12 +146,12 @@ export class OwnDatatableComponent implements OnInit, AfterViewChecked, OnDestro
   displayMode = 'medium';
   defaultSort: any;
 
-  usersId = '';
+  usersId?: string;
   usersIdSub!: Subscription;
 
   subscriptionDeleteOwnDocument!: Subscription;
 
-  downloadedDocumentId = '';
+  downloadedDocumentId?: string;
   fileType = 'csv';
 
   _columns = ['dateEdited', 'dateObserved', 'locality', 'taxon', 'gatheringsCount', 'unitCount', 'observer', 'form', 'id', 'publicityRestrictions'];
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/shared-modules/technical-news/technical-news-dumb/technical-news-dumb.component.ts
```diff
@@ -9,14 +9,24 @@ import { News } from '../../../shared/model/News';
...
 export class TechnicalNewsDumbComponent {
   @Input() set news(news: News[] | null) {
-    this.technicalNews = news?.filter(newsItem => {
-      const days = 1;
-      const isNew = Date.now() - parseInt(newsItem.posted, 10) < (days * 86400000); // number of milliseconds in a day
-      const isTechnical = newsItem.tag === 'technical';
-      return isTechnical && isNew;
-    });
+    if (news) {
+      this.technicalNews = news?.filter(newsItem => {
```
aorin, line 13: This could be changed to `this.techicalNews = (news || []).filter(...`, then the if clause can be removed and the type can be array (`technicalNews: News[] = [];`)

projects/laji/src/app/shared-modules/own-submissions/own-datatable/own-datatable.component.ts
```diff
@@ -132,31 +137,31 @@ export class OwnDatatableComponent implements OnInit, AfterViewChecked, OnDestro
...
-  usersId: string;
-  usersIdSub: Subscription;
+  usersId = '';
+  usersIdSub!: Subscription;
 
-  subscriptionDeleteOwnDocument: Subscription;
+  subscriptionDeleteOwnDocument!: Subscription;
 
-  downloadedDocumentId: string;
+  downloadedDocumentId = '';
```
aorin, line 154: Instead of setting a default value, you could make the variable optional `downloadedDocumentId?: string` or set the type as `downloadedDocumentId: string | undefined`

aorin (approved):
Looks fine to me. In the future, it might be better/easier to just change the types and not any values. A change like `year: string` -> `year = ''` could possible break something, for example if there was a check like `if (year === undefined)` somewhere. But I didn't see anything like that here

PR #678, opened 2024-11-21
Filter for local status 187438532

MLumme:
Old pivotal task here: https://www.pivotaltracker.com/story/show/187438532
Testable here: https://187438532.dev.laji.fi/observation/list

Added new Muuttaja-filter and info text under Yksilö to use local filter for individuals, translations are waiting in branch crowdin.


rpulkka (approved):
Good!

PR #679, opened 2024-11-21
Dont require mandatory columns when deleting in excel import 183507263

MLumme:
Old pivotal task here: https://www.pivotaltracker.com/story/show/183507263
Testable here: https://183507263.dev.laji.fi/

Bypassed excel import mandatory column validation inside the excel importer if the row is being deleted.


rpulkka (approved):
OK!

PR #701, opened 2024-12-18
Refactor informal taxon groups usage according to api changes

### LINKED ISSUE #700: Update `/informal-taxon-groups` usage according to API changes
See the parent task for the reasoning.

Only the `/parents/` -> `/parent` needs to be implemented. The current implementation for `/tree`, `/roots`, `/:id/siblings` assume already that the result fits in a single page, and the signature is compatible as the API returns the arrays wrapped in ` { results }`

olzraiti:
* Use `informal-taxon-groups/:id/parent` instead of `informal-taxon-groups/:id/parents` (reasoning in
  https://github.com/luomus/laji-api/issues/51)
* Refactor informal taxon groups client
* Fix null pointer error in `informal-list-breadcrumb` (currently broken in https://beta.laji.fi/taxon/browse?informalGroupFilters=MVL.342 - open the console and there are errors spitted)

Closes #700

### DIFF
--- a/projects/laji/src/app/shared/model/ArrayResult.ts
+++ b/projects/laji/src/app/shared/model/ArrayResult.ts
@@ -1,0 +1,4 @@
+
+export interface ArrayResult<T> {
+  results: T[];
+}
--- a/projects/laji/src/app/shared/service/util.service.ts
+++ b/projects/laji/src/app/shared/service/util.service.ts
@@ -71,7 +71,7 @@ export class Util {
   }
 
   /**
-   * Remove undefined values from object and any key specified by the key parameter
+   * Remove all undefined and null values from object
    *
    * @param obj object to remove keys from
    * @param keys array of keys that should be removed
--- a/projects/laji/src/app/shared/service/util.service.ts
+++ b/projects/laji/src/app/shared/service/util.service.ts
@@ -2,6 +2,10 @@ import { NavigationEnd, Event } from '@angular/router';
 import * as merge from 'deepmerge';
 import { Document } from '../model/Document';
 
+export type WithNonNullableKeys<T, K extends keyof T> = T & {
+  [P in K]-?: NonNullable<T[P]>;
+};
+
 export class Util {
   /**
    * Clones the object using JSON stringify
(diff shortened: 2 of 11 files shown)


projects/laji/src/app/shared/model/ArrayResult.ts
```diff
@@ -0,0 +1,4 @@
+
+export interface ArrayResult<T> {
+  results: T[];
```
Blodir, line 3: This api is a bit strange. Why not return a simple array from the api?

projects/laji/src/app/shared/service/util.service.ts
```diff
@@ -66,6 +70,22 @@ export class Util {
     }, {} as Partial<T>);
   }
 
+  /**
+   * Remove all undefined and null values from object
+   *
+   * @param obj object to remove keys from
+   * @param keys array of keys that should be removed
+   */
+  public static withNonNullableKeys<T extends {[prop: string]: any}, K extends keyof T>(obj: T): WithNonNullableKeys<T, K> {
```
Blodir, line 79: Shouldn't it be `removeNullableFields` or at least `withNoNullableKeys`? This name sounds to me like you are adding extra keys rather than removing nullable keys

Blodir: another alternative: `getNonNullableKeys`

Blodir (changes requested):
Lint job seems to have some issue.. I'll try rerun

projects/laji/src/app/shared/model/ArrayResult.ts
olzraiti: So we can give JSON-LD context to the result, without needing each item to have the context: https://w3c.github.io/json-ld-bp/#top-level-object

projects/laji/src/app/shared/service/util.service.ts
olzraiti: The term `NonNullable` comes from TypeScript: [NonNullable](https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype)

Though the "keys" word usage is weird... How about `withNonNullableValues`? `getNonNullableKeys` sounds to me like it would return keys of an object.

olzraiti: I pushed `withNonNullableValues` as a proposal. Updated the JS doc comment also. It was copy pasted from some other utility fn so there was an extra param.

Blodir: I still don't think I would immediately understand what `withNonNullableValues` does without reading the comment. Usually the word `with` I would associate with adding something to the result, eg. `withLatestFrom` in rxjs. Maybe `excludeNullables` or invert it `selectNonNullables`.

Sorry this is getting rather pedantic... I'll give the PR an approve.

Blodir (approved):

PR #712, opened 2025-01-13
Remove document batch upload checking job status documents

olzraiti:
Removes usage of the job status `documents` property. Currently it seems to try to validate the API response that it has correct length for the documents. We should just trust the API for this, no need to carry the documents over the job status check.


aorin (approved):
Looks ok

Issue #711, opened 2024-11-15
Remove .shp file GIS download option and deprecation notfication


wkmor1:
- [x] From UI for Citable files
- [x] From UI for Pyha
- [x] From Lite download UI

aorin:
@wkmor1 I removed the option from all places, ready to go to the next release after approved

https://711.dev.laji.fi/observation/list?target=MX.37623
https://711.viranomaiset-dev.laji.fi/usage/my-downloads
https://staging-pyha.laji.fi/

aorin:
@wkmor1 updated pyha, this task should be done now

PR #723, opened 2025-01-28
Remove shp file format from gis downloads 

aorin:
#711 

https://711.dev.laji.fi/observation/list?target=MX.37623
https://711.viranomaiset-dev.laji.fi/usage/my-downloads


projects/laji/src/app/shared-modules/download-request/file-download.service.ts
```diff
@@ -21,7 +21,7 @@ export enum FileType {
 @Injectable({providedIn: 'root'})
 export class FileDownloadService {
   fileType: FileType = FileType.standard;
-  format: FileFormat = FileFormat.shp;
+  format: FileFormat = FileFormat.gpkg;
```
rpulkka, line 24: File download service uses gpkg now that shp was removed?

projects/laji/src/i18n/fi.json
```diff
@@ -439,7 +438,6 @@
   "friend.requests": "Kaveripyynnöt",
   "friend.requestSent": "Kaveripyyntö lähetetty",
   "gathering.section.outsideSection": "Ei osiota",
-  "geoConvert.error.too_complex": "Joidenkin havaintojen geometriat ovat liian monimutkaisia .shp-tiedostomuodolle (GeometryCollection). Käytä sen sijaan .gpkg-muotoa tai valitse toinen geometriatyyppi.",
```
rpulkka, line 442: Did you delete references of "too_complex"?

rpulkka (commented):
Can you confirm the two comments I made ^

projects/laji/src/app/shared-modules/download-request/file-download.service.ts
aorin: Yeah, gpkg is the only gis download format now

projects/laji/src/i18n/fi.json
aorin: I had forgotten to delete it, deleted it now

rpulkka (approved):
Looks good!

PR #724, opened 2025-01-28
Add OGC API theme page to PAP

aorin:
Can be tested here: https://18812669.viranomaiset-dev.laji.fi/theme/ogcapi

Adds a theme page where you can order an api key for the OGC API.

closes luomus/laji-pygeoapi#6


MLumme (commented):
LGTM

PR #725, opened 2025-01-28
Excel import named place

### LINKED ISSUE #568: Link imported documents to named place
On Vihko, when user is importing documents via Excel import to a census that has named places, and user has selected that named place on the Excel file, the resulting documents should be linked to the named place.

Get coordinates and locality names from the named place, so that they appear correctly on Laji.fi (unclear how this is now). If user has provided both named place and locality & coordinates, ignore the locality & coordinates.

aorin:
closes #568 

https://568.dev.laji.fi/project/MHL.1136/import

The excel import augments the documents with named place data if they contain named place ids. This previously worked for `gathering`-level named places, now it also works for `document`-level places.


olzraiti (approved):
LGTM

Issue #714, opened 2025-01-17
Double check that SSR doesn't make not-needed DW queries


esko-piirainen:
Reduce number of DW queries (calls to:  grapql or warehouse directly)
Also finbif-bib-top is queried a lot, that is not needed for SSR either

- [x] frontpage
- [x] species cards
- [x] occurrence search "frontpage"
- [x] others pages that make dw queries??

aorin:
@esko-piirainen I disabled all the queries to `/api/graphql` and `/api/warehouse` in the server. Also the publications section is hidden for SSR. Can be tested here https://714.dev.laji.fi/ but the SSR is only enabled in beta.laji.fi

esko-piirainen:
@aorin When I got to https://laji.fi/observation  (main page of occurrence search), it does all the queries in the image
Does it do any of these for SSR?
![Image](https://github.com/user-attachments/assets/9636bae8-0d9e-4fc4-af81-f4c0b1151f3f)

At least those /aggregate queries are not needed for SSR (x2)


Not sure how to test on that dev branch?

esko-piirainen:
I found the commit... Looks like except for finbib-bib-top you put it quite deep, so that no queries should be made no matter what the source is.  Still, can't really test this?

aorin:
It can't be tested (except locally) before it's in the beta. After it's in the beta, the server side rendered page can be viewed by disabling the javascript and it can be compared with the production version to see what parts are missing.

esko-piirainen:
Good! Lets approve this when in beta...

esko-piirainen:
Looks good in beta!

PR #726, opened 2025-01-28
Reduce ssr dw queries

aorin:
#714

Reduces the number of queries in ssr to data warehouse by
- removing all graphql queries in the server
- removing all queries to `warehouse` endpoint in the server (most were already removed)
- hides the publications section in the front page in the server


MLumme (commented):
LGTM

PR #727, opened 2025-01-28
Bsg small changes

aorin:
Can be tested here: https://bsg.laji.fi

Small changes:
- adds a frog taxon type
- changes default sample rate for bat recordings
- adds a sound type select


MLumme (approved):
LGTM

Blodir (approved):

PR #746, opened 2025-02-24
Generate trait search filters

Blodir:
- `scripts/trait-search-codegen.ts` generates filters and table columns for trait db trait search by traversing the typescript AST of the api type definition file. I couldn't easily traverse the swagger directly, because `swagger-parser` threw errors on our openapi schema. Besides, using the typescript compiler API does make sense since it's used later to generate the filters/columns anyway.
- added `lu-form-primitive-list` (replaces string-list). It's used for form input of simple lists. String/number/enum atm. Could be expanded to other form input types such as `date` as needed.
- trait search now uses the primive-list component for all of the array type filters


projects/laji-ui/src/lib/forms/primitive-list/primitive-list.component.ts
```diff
@@ -0,0 +1,98 @@
...
+  ],
+  changeDetection: ChangeDetectionStrategy.OnPush
+})
+export class FormPrimitiveListComponent implements ControlValueAccessor {
+  @Input() inputId: string | undefined;
+  /**
+   * If kind === 'enum', then enumVariants must be defined.
+   */
+  @Input() kind: Kind = 'string';
+  @Input() enumVariants!: ExtractKind<this> extends 'enum' ? string[] : undefined;
```
aorin, line 28: The types could be simplified here too. If the class is made generic `FormPrimitiveListComponent<T extends Kind>`, then all the types would look nicer
```
@Input({ required: true }) kind!: T;
@Input() enumVariants!: T extends 'enum' ? string[] : undefined;
```

The only downside is that the `kind` input couldn't have a default value so easily then but it doesn't really need it?

aorin (changes requested):

projects/laji-ui/src/lib/datatable/datatable.component.html
```diff
@@ -42,14 +42,14 @@
         <td *ngFor="let colIdx of selectedColumns" #rowRef>
           <ng-container *ngIf="columnHasTemplate(columns[colIdx]); else defaultTemplate">
             <ng-container
-              *ngTemplateOutlet="columns[colIdx]['cellTemplate'];
-              context: columns[colIdx].prop
-                ? { $implicit: row, prop: getNestedValue(row, columns[colIdx]['prop']) }
+              *ngTemplateOutlet="$any(columns[colIdx])['cellTemplate'];
```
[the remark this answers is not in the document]
Blodir: `SortableColumnWithTemplateButNoProp` covers the following issue: if there's a sortable column with no `prop` it can still be valid as long as it has a `cellTemplate` and `sortFn`. Eg. if there was only `ColumnWithTemplate` then it would have to enforce the existence of `prop` to ensure there's no invalid inputs and you would have cases where prop has to be included even if it's never used. It wouldn't be a big deal to have the extra prop in there, but idk. In the end I think it comes down to preference and I personally favor precision/irreducibility and type-safety over readability 🤷‍♂️ Eg. if we had just a `DatatableColumn { prop?, sortable, sortFn?, cellTemplate? }` type, it would be readable sure, but it would allow the user to give incorrect inputs that they only notice during runtime (i.e. bugs).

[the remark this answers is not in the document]
Blodir: Super weird how the `as` keyword works as long as it's not in combination with the bracket notation... but I was able to get rid of the template $anys, thanks! I did also have to replace the `ngIf else` with two ifs to have type narrowing work but I guess that's fine

projects/laji-ui/src/lib/forms/primitive-list/primitive-list.component.ts
Blodir: Unfortunately angular does not respect generics in components :/ it's weird because I have found many references of people using them. Maybe in a newer version angular does some magic? Anyways, angular treats generic components as just `any`, so the component would be seen as `FormPrimitiveListComponent<any>` by angular, effectively breaking type checking for the inputs. I just tested this to make sure. My solution of avoiding generics works on the other hand eg. 
image

Using generics angular templates would allow `[enumVariants]` to be present when it shouldn't be

Blodir: This line
```
    [type]="{ 'string': 'text', 'number': 'number', 'enum': 'hidden' }[kind]"
```
in `primitive-list.component.html` also breaks using generics, because angular thinks that `kind` is `any`

aorin: The generics should work, Angular can infer the type of the component from the input. I tested it and works here without problems. It gives the error from `enumVariants` if the `kind` is `string` as expected and that line above also works. Maybe you had some other error there or your ide is not up-to-date if it showed you the error?

Blodir: Hmm maybe it is a language service problem... the angularLS is quite buggy with the laji repo for me in general as well. The generic version errors using `npm run build:ssr:local` as it should:

```
export class FormPrimitiveListComponent<K extends Kind, T extends KindToConcreteType<K>> implements ControlValueAccessor {
  @Input() inputId: string | undefined;
  /**
   * If kind === 'enum', then enumVariants must be defined.
   */
  @Input() kind!: K;
  @Input() enumVariants!: K extends 'enum' ? string[] : undefined;

  valueList: T[] = [];
  onChange: (value: T[]) => void = () => {};
  onTouched: (value: T[]) => void = () => {};

...
```

```
    [type]="{ 'string': 'text', 'number': 'number', 'enum': 'hidden' }[kind]"
```

With erroneous input (give enumvariants with kind=string) and `npm run build:ssr:local`.

Both solutions error with:
Error: projects/laji/src/app/+trait-db/trait-db-datasets/trait-db-dataset-editor/trait-db-dataset-editor.component.html:64:76 - error TS2322: Type 'string[]' is not assignable to type 'undefined'.

64     <lu-form-primitive-list inputId="additionalIdentifiers" kind="string" [enumVariants]="['a', 'b']" formControlName="additionalIdentifiers"></lu-form-primitive-list>

Strangely I have another problem:
```
"Element implicitly has an 'any' type because expression of type 'any' can't be used to index type '{ string: string; number: string; enum: string; }'."
```
AngularLS gives me an error here too, but ssr build doesn't...

Blodir: I'll use the generic solution since it's just a language service issue it seems

Blodir:
Strict templates in the angular build just caught a bug, hooray! In datatable component, the `getNestedValue` function was assuming that a prop is a string, when it could be any `Keyable`. Fixed!

projects/laji-ui/src/lib/forms/primitive-list/primitive-list.component.ts
aorin: I've also had problems with the webstorm giving wrong errors but nowadays it works mostly fine. The types look good but could you still make the `kind` input `required` as using the `!` mark without the `required` attribute can cause problems as the compiler won't complain if it's missing

projects/laji-ui/src/lib/datatable/datatable.component.html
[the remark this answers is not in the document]
aorin: Hmm shoudn't the `SortableColumnWithTemplateButNoProp`'s `sortFn` at least be the same type as `BasicColumn`? Like isn't the input always same? Now that I look the place where it's used, there's something weird there as well since there's two times the check `col.sortFn ? col.sortFn` 🤔
```
      const sortFn = col.sortFn ? col.sortFn
        : (a: DatatableRow<RowProp>, b: DatatableRow<RowProp>) =>
          col.sortFn
            ? col.sortFn(a, b)
            : a[col.prop] > b[col.prop] ? 1 : a[col.prop] < b[col.prop] ? -1 : 0;
```

The template looks nice without the `$any`s. Yeah the type narrowing has some limits at the moment, I'm wondering if the types work better with the new angular control flow, we could test it someday

projects/laji-ui/src/lib/forms/primitive-list/primitive-list.component.ts
Blodir: good point, fixed!

projects/laji/src/app/+trait-db/shared/trait-search/trait-search-filters/trait-search-filters.component.html
```diff
@@ -4,15 +4,15 @@
   <div class="mb-5 d-flex flex-gap-4">
     <div class="dataset-filter flex-grow">
       <label for="trait-search-filters-dataset">Dataset</label>
-      <select class="form-control" id="trait-search-filters-dataset" [formControl]="form.get('dataset')">
+      <select class="form-control" id="trait-search-filters-dataset" [formControl]="$any(form.get('dataset'))">
```
rpulkka, line 7: What happens, if you don't add the $any? And can it be any other type than $any?

scripts/trait-search-codegen.ts
```diff
@@ -0,0 +1,203 @@
...
+};
+
+const generateDatatableColumns = (tree: any) => {
+  const cols = accDatatableColumnsRecursive(tree);
+  console.log(cols);
+  generateExportObjectLiteral(cols, 'cols', './projects/laji/src/app/+trait-db/shared/trait-search/trait-search-table-columns.ts');
+};
+
+const generateFilters = (tree: any) => {
+  const cols = accDatatableColumnsRecursive(tree);
```
rpulkka, line 147: Looks like `accDataTableColumnsRecursive` is being called first on `generateDatatableColumns` and then again on `generateFilters`. Would it be more effecient, if you would call it on main and give `cols` as param to the two functions?

projects/laji-ui/src/lib/datatable/datatable.component.html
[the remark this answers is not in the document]
Blodir: idk how the double check for sortFn ended up there. Fixed. Also extracted the SortFn into an utility type so now they are identical

aorin (approved):
:+1: Looks good

scripts/trait-search-codegen.ts
Blodir: Performance doesn't matter here since it's not a script that is run regularly, but I fixed it anyway

rpulkka (approved):
Well done!

Issue #751, opened 2025-01-28
Improve PAP backend


olzraiti:
The PAP has all kind of weird bugs, that stem from weirdness in the PAP backend.

Instead of killing these bugs one by one, we should aim to fix the design flaws in the PAP backend instead.

Examples:

1. a request with response status `403` trigger a page reload.
2. The allowed queries hash table for graphql (and possibly other queries)

I think we could start by evaluating the code together, then decide what security solutions are actually relevant to the use case of PAP and create a development plan.

This Epic was created based on this Slack discussion: https://luomus-ict.slack.com/archives/C01515GSM6V/p1738061272299669

esko-piirainen:
Next up: still needs some tests, then should be ready for  staging/prod. Meeri works on this the coming sprint.

PR #757, opened 2025-03-10
Fix PAP redirecting

aorin:
#751
https://751.viranomaiset-dev.laji.fi/
(related to https://github.com/luomus/viranomaiset-backend/pull/1)

Previously PAP used to redirect to login if it got any `401` or `403` response, now it redirects only if it gets `401` with a `'Login Required'` message. I removed the `forceLogin` environment variable since it was only used with this and it wasn't very descriptive. I moved the code from shared modules to vir instead.

I also removed `build:vir:integrate-w-backend:watch` script and added a proxy config instead. I don't see a reason why the frontend would be developed without the backend as it has quite many custom endpoints. The updated development instructions are [here](https://wiki.helsinki.fi/xwiki/bin/view/luomusict/Luomus%20Biodiversity%20Informatics%20Unit/IT%20Systems/viranomaiset.laji.fi/).


olzraiti (approved):
LGTM!

PR #805, opened 2025-06-30
Fix expired person token error handling

Blodir:
feature branch: https://536465345673.dev.laji.fi/
(not yet approved)


rpulkka (approved):
OK!

Issue #812, opened 2025-08-08
Update download citation page


esko-piirainen:
https://dev.laji.fi/citation/HBF.12916

- [ ] readme.txt name has changed
- [ ] instructions on this page to point to info.laji.fi ?

aorin:
Can be tested here https://812.dev.laji.fi/citation/HBF.12916, the license name is added to the translations so it can later be translated in Crowdin. The links were already in Crowdin

esko-piirainen:
Looks ok. Must remember to add those texts when in beta...

PR #823, opened 2025-09-01
License file name to translations

aorin:
https://github.com/luomus/laji/issues/812
https://dev.laji.fi/citation/HBF.12916

### DIFF
--- a/projects/laji/src/app/shared-modules/download-request/download-request/download-request.component.html
+++ b/projects/laji/src/app/shared-modules/download-request/download-request/download-request.component.html
@@ -14,7 +14,7 @@
         role="alert"
         class="alert alert-success"
         [innerHTML]='"downloadRequest.fileDownload.licenseReminder" | translate:{
-          filename: "<code>readme.txt</code>"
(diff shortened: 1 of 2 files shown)


projects/laji/src/app/shared-modules/download-request/download-request/download-request.component.html
```diff
@@ -14,7 +14,7 @@ <h4>{{ 'downloadRequest.download' | translate }}</h4>
         role="alert"
         class="alert alert-success"
         [innerHTML]='"downloadRequest.fileDownload.licenseReminder" | translate:{
-          filename: "<code>readme.txt</code>"
+          filename: "<code>" + ("downloadRequest.fileDownload.licenseFile" | translate) + "</code>"
```
Blodir, line 17: Could this be refactored to not use innerHTML
eg.
```html
<div ...>
  <code>
    {{ getLicenseReminderText$ | async }}
    (or using translate.instant without the async is fine too)
  </code>
</div>
```

InnerHTML behavior is pretty hard to predict... i'm surprised that pipes work within innerhtml at all actually

Blodir (changes requested):

PR #828, opened 2025-09-09
Autocomplete against api v1

olzraiti:
All autocomplete queries are done with the new API client. In addition, I made the API client filter undefined query params.

Here are the relevant breaking changes of the new API that are taken into account:

### `/autocomplete/person`

* renamed as `/autocomplete/persons`
* results are wrapped in `results`
* query param `q` renamed as `query`
* query param `limit` renamed as `pageSize` (supports pagination now)
* query param `includePayload` is removed, fields can be filtered with `selectedFields` instead

### `/autocomplete/friends`

* results are wrapped in `results`
* query param `q` renamed as `query`
* query param `limit` renamed as `pageSize` (supports pagination now)
* query param `includePayload` is removed, fields can be filtered with `selectedFields` instead
* query param `includeSelf` is removed, self is always included

### `/autocomplete/taxon`

* renamed as `/autocomplete/taxa`
* results are wrapped in `results`
* query param `q` renamed as `query`
* in the result, the `taxonRankId` renamed as `taxonRank`
* in the result, the `informalTaxonGroups` renamed as `informalGroups`
* in the result, the `matchType` renamed as `type`
* query param `includePayload` is removed, fields can be filtered with `selectedFields` instead

### DIFF
--- a/projects/laji-api-client-b/src/laji-api-client-b.service.ts
+++ b/projects/laji-api-client-b/src/laji-api-client-b.service.ts
@@ -125,7 +125,16 @@ export class LajiApiClientBService {
   }
 
   private getRequestOptions(queryParams: any, requestBody: any, lang: string) {
-    return { params: queryParams, body: requestBody, headers: {
+    return {
+      params: Object.keys((queryParams || {})).reduce((filteredQueryParams, key) => {
+        const param = (queryParams || {})[key];
+        if (param === undefined) {
+          return filteredQueryParams;
+        }
+        filteredQueryParams[key] = queryParams[key];
+        return filteredQueryParams;
+      }, {} as any),
+      body: requestBody, headers: {
 			'API-Version': '1',
 			'Accept-Language': lang
 		} };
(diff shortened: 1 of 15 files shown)


projects/laji-api-client-b/src/laji-api-client-b.service.ts
```diff
@@ -125,7 +125,16 @@ export class LajiApiClientBService {
...
-    return { params: queryParams, body: requestBody, headers: {
+    return {
+      params: Object.keys((queryParams || {})).reduce((filteredQueryParams, key) => {
+        const param = (queryParams || {})[key];
+        if (param === undefined) {
+          return filteredQueryParams;
+        }
+        filteredQueryParams[key] = queryParams[key];
+        return filteredQueryParams;
+      }, {} as any),
```
aorin: Could have a little stricter type like Record<string, any>

aorin (approved):
:+1:

Issue #672, opened 2024-11-08
Add printing functionality to laji.fi observation map


AlpoTurunen:
Printing observations and their locations directly from the laji.fi map view would be good for users who don't know how to use GIS tools. That functionality has already been implemented in [laji.fi/map](http://laji.fi/map) but not yet for species occurrences.

wkmor1:
Slack discussion https://luomus-ict.slack.com/archives/CQEQSQCP7/p1731068494301409

esko-piirainen:
If easy to do (consult Olli), lets add the icon to obs search map!

PR #833, opened 2025-10-02
Observation map print

aorin:
#672 
https://672.dev.laji.fi/observation/map

Adds a print pdf button to the observation map.

This task had a few challenges. The first one was that the pdf styles were broken on localhost because the stylesheet is a link to a local file (`<link href="styles.css">`). I fixed it so that it so that it loads the stylesheet and replaces the link with the actual styles.

The second quite big problem was that the bootstrap has a rule that sets background color and box shadow to transparent when printing. So all the map circles were missing a color. I fixed the problem here by adding a class name for each of the elements and setting the print styles in `projects/laji/src/app/shared-modules/observation-map/observation-map/observation-map.component.scss`. If we are going to implement more printing features, then we might have to consider some other solution since it takes some work to add the styles individually.


aorin:
It doesn't seem easy to remove the style on the frontend side (that method can't be used since javascript is not executed on the html-to-pdf endpoint). But since we don't need that style anywhere I though the easiest option is to download the bootstrap with the print styles disabled. The version that was available was 3.4.1, we had previously 3.3.7 but there shouldn't be any breaking changes. The colors still didn't appear without the `print-color-adjust: exact;` rule, I added it as a global rule when printing since we probably want the colors to look the same on default

PR #418, opened 2023-06-22
Section for complete lists [#185176715]

rpulkka:
https://185176715.dev.laji.fi/save-observations
https://www.pivotaltracker.com/n/projects/2346653/stories/185176715

This pull request adds new section for complete list forms to the save-observations page. 

NOTE (also documented in the Next Update document): When publishing to production, remember to change all BIOMON forms' category to 'MHL.categoryBiomonCompleteLists' on Laji-Form. All BIOMON forms inherit the Dragonfly form (MHL.932) so change to that form might be inherited to all of the other forms, check the Buttefly form (MHL.1040) to make sure that the change was inherited.

PR #835, opened 2025-10-02
Implement MHL.hideFromSidebar [closes #418]

olzraiti:
Also include `reserve` property always when named places. This is because the line transect "Vakiolinjat" form is now a sub form of top level form that holds the restriction options, targetting the top collectionID. The sub form MHL.1162 shouldn't have the restriction options, because then it wouldn't inherit the top level form collections form permissions.


rpulkka (approved):
OK!

PR #836, opened 2025-10-10
LajiApiClientB handles person token

olzraiti:
`LajiApiClientB` handles injecting person token to requests globally.

`/person` and `/authentication-event` (used to be `/person-token`) are implemented with LajiApiClientB

### DIFF
--- a/projects/kerttu-global/src/app/+identification/expertise/expertise.component.ts
+++ b/projects/kerttu-global/src/app/+identification/expertise/expertise.component.ts
@@ -1,6 +1,5 @@
 import { Component, OnInit, ChangeDetectionStrategy, ChangeDetectorRef } from '@angular/core';
 import { Observable, Subscription, forkJoin } from 'rxjs';
-import { UserService } from '../../../../../laji/src/app/shared/service/user.service';
 import { Profile } from '../../../../../laji/src/app/shared/model/Profile';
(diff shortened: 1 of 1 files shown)


projects/kerttu-global/src/app/+identification/expertise/expertise.component.ts
```diff
@@ -95,7 +95,7 @@ export class ExpertiseComponent implements OnInit {
     this.profile!.birdwatchingActivityLevel = this.birdwatchingActivityLevel as BirdwatchingActivityLevelEnum;
     this.profile!.birdSongRecognitionSkillLevels = this.birdSongRecognitionSkillLevels;
 
-    return this.personService.personUpdateProfileByToken(this.profile!, this.userService.getToken()).subscribe(() => {
+    return this.api.put('/person/profile', undefined, this.profile! as any).subscribe(() => {
```
aorin: Something weird is happening with the types. The new api client is supposed to have the correct types so there probably shouldn't be need to use `any`. In the create profile method the type of the body is `components["schemas"]["Profile"]` and in the update profile method the type is `components["schemas"]["profile"]`. Then the `Profile` type has
`settings: {[key: string]: unknown;}` and the `profile` type has `settings?: Record<string, never>;`. Why are they different and the `Record<string, never>` doesn't look correct since it's a type for empty object.

aorin (approved):

PR #838, opened 2025-10-14
Trait db improvements

Blodir:
- Refactored trait search filters to use (more) "reactive forms"
- Changed taxon rank filter to from many textfields for each rank to rank enum + search textfield
- Added datasetId param to trait search filters, and added the search to dataset pages. The datasetId doesn't show in query params if it's set by the @Input property

### DIFF
--- a/projects/laji/src/app/+trait-db/shared/trait-search/trait-search-filters/additional-filters.component.ts
+++ b/projects/laji/src/app/+trait-db/shared/trait-search/trait-search-filters/additional-filters.component.ts
@@ -116,7 +116,7 @@ export class TraitSearchAdditionalFiltersComponent implements ControlValueAccess
         Object.entries(values)
           .reduce((p, [k, v]) => {
             if (values[k]) {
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/+trait-db/shared/trait-search/trait-search-filters/additional-filters.component.ts
```diff
@@ -81,62 +81,84 @@ export type AdditionalFilterValues = Partial<{
...
-  ngOnInit(): void {
-    if (this.initialValue) {
-      const propsMappedToFormKeys: any = {};
-      Object.entries(this.initialValue).forEach(([k, v]) => {
+  ngOnDestroy(): void {
+    this.subscription.unsubscribe();
+  }
+
+  writeValue(obj: any): void {
+    const propsMappedToFormKeys: any = {};
```
aorin: `any` types could be replaced with a stricter type if possible, this could probably be `Record<string, any>`?

aorin (changes requested):

aorin (approved):
Looks good!

Issue #596, opened 2024-10-18
Handle UTF-8 without BOM on Vihko import


mikkohei13:
Vihko import seems to expect UTF-8 with BOM. Make it handle file without BOM as well.

Attached is a test file as UTF-8 without BOM.

[Import_UTF8_wo_BOM.csv](https://github.com/user-attachments/files/17431974/Import_UTF8_wo_BOM.csv)

PR #844, opened 2025-10-14
Import handles utf-8 csv files

aorin:
https://596.dev.laji.fi/vihko/tools/import
https://github.com/luomus/laji/issues/596

Fix import didn't work with utf-8 files.


rpulkka (approved):
OK!

PR #885, opened 2026-01-09
Archipelago bird census named place selection [Closes #433]

rpulkka:
https://github.com/luomus/meta/issues/433
https://433.dev.laji.fi/project/MHL.1166/form/places

Adds new query params `filterBy` and `tab` to /form/places.

In np-list component, visible values are used in initData to fix filtering to use enum values instead of enum keys.

datatable.component required changes to avoid scrolling before it is visible. The component also emits filteredIDs based on filtered rows.

Filter was previously only applied to the list tab, but now places are also filtered on the map tab. This includes transforming between `fullIdx` and `filteredIdx` in np-map.component, as the map uses index to refer to the active place and the map indexing changes when a filter changes the number of the shown places.


olzraiti (changes requested):
One minor whining, otherwise looks good!

PR #910, opened 2026-02-24
History service improvements [Closes #908]

### LINKED ISSUE #908: Fix np-selection not returning to the selected named place
When going to named place selection via a link where a named place is in the query params https://beta.laji.fi/project/MHL.1166/form/MHL.1166/places?activeNP=MNP.56680 opening the form and coming back results to no named place being selected.

rpulkka:
Closes: https://github.com/luomus/laji/issues/908
Branch: https://807.dev.laji.fi/project/MHL.1166/form/MHL.1166/places?activeNP=MNP.56680

On addresses with replaceUrl (meaning that only the latest query param change should be saved to history) there was an issue that the first visit to the replaceUrl didn't get saved. So reload on a replaceUrl page with params -> going to another page -> coming back would lose query params and now it doesn't because of this if case: `if (this.router.currentNavigation()?.extras.replaceUrl)`

Also, the confusing `hasBack` variable was replaced with `isInitialLoad` to check the situation where user loads the first page on the first time. Note that /user, /login and /logout paths are also now considered in `isInitialLoad` which makes home-data.service use cache-first on them as well, unlike the earlier no-cache. `getPrevious` was deleted as we can use `browserService.goBack` in `onDocumentDeleted`.

### DIFF
--- a/projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
+++ b/projects/laji/src/app/shared-modules/document-viewer/document/document.component.ts
@@ -29,7 +29,7 @@ import { TemplateForm } from '../../own-submissions/models/template-form';
 import { Router } from '@angular/router';
 import { LocalizeRouterService } from '../../../locale/localize-router.service';
 import { DeleteOwnDocumentService } from '../../../shared/service/delete-own-document.service';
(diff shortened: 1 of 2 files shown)


MLumme (approved):
LGTM

PR #913, opened 2026-02-27
Initialize lang at app initialization

olzraiti:
Currently lang is initialized by `LocalizeGuard` which executes very late as it needs to route tree or something. It takes ~2s on my machine on http://localhost:3000/en/taxon/MX.34567.

This PR makes lang initialization happen during app initialization.


Blodir:
It feels hacky to opt out of angular router like this and use window directly instead... shouldn't localizeguard always be executed before instantiating any components anyways? Like angular needs to know what route to render, it gets that from resolving routes and hits LocalizeGuard on the way. Does this fix any practical issues?

Maybe some services that need the lang may be launched before route resolution, like BaseDataService 🤔

Issue #881, opened 2025-12-16
DNA/Sequence data handling


mikkohei13:
As someone working on DNA sequence data, I want to be able to save it to Vihko, so that 
* It is safely stored
* It is available for others to use

- [x] Add field to Vihko
- [x] Make occurrences filterable by absence/presence of sequence text
- [x] add sequence to List -tab on browse occurrences

Planning as part of Funtaxa in 2/2026?

rpulkka:
Waiting for next release.

PR #925, opened 2026-03-10
Dna seq filter 881

esko-piirainen:
Adds DNA section to occurrence search with one filter in the section
Adds DNA sequence as selectable field to list ( + lightweight downloads)

Note: Did not run tests
Note: Did not add the new DW filter  (hasSequenceText=true|false)  to laji-api-client  (seems not to use the client yet?)


Blodir:
#881

MLumme (approved):
LGTM

PR #939, opened 2026-03-23
Api client refactors

Blodir:
I had to do a few temporary unsafe type casts to limit the scope of the changes

Also had to stick to the old api for `api/warehouse/query/sample/aggregate` since that path appears to not exist in the new api yet


projects/laji/src/app/shared-modules/document-viewer/document-viewer.facade.ts
```diff
@@ -85,25 +88,29 @@ export class DocumentViewerFacade {
...
     this.updateState({..._state, showModal: true});
-    this.documentApi.findById(param.document, this.userService.getToken()).subscribe((document) => {
+    this.api.get('/documents/{id}', { path: { id: param.document } }).subscribe(document => {
       this.showDocument({
         ...param,
         document
       });
     });
+    this.documentApi.findById(param.document, this.userService.getToken()).subscribe((document) => {
+    });
```
olzraiti: forgot to remove?

olzraiti (changes requested):

olzraiti (approved):

PR #946, opened 2026-03-30
Api client migrations

Blodir:


projects/laji/src/app/shared/service/triplestore-label.service.ts
```diff
@@ -74,16 +67,15 @@ export class TriplestoreLabelService {
...
-                includePublic: true
-              },
-              '1',
-              '1'
-              ).pipe(
+                includePublic: true,
+                page: 1,
+                pageSize: 1
+              }
+            } as any).pipe(
```
Blodir: the /named-places query shouldn't require collectionId? Old one seems to be optional at least. This any can be removed once collectionId is made optional again @olzraiti

olzraiti: fixed that

projects/laji/src/app/+observation/main-result/main-result.service.ts
olzraiti, line 1: Looks like the whole service can be removed? It's a provider only for `main-result.component.ts` but it doesn't ever use the service

olzraiti (changes requested):

projects/laji/src/app/+observation/main-result/main-result.service.ts
Blodir: ahhhh... i should check more carefully before refactoring... Yeah looks like it can be just removed

PR #953, opened 2026-04-08
Small refactor to label designer to get rid of old laji api client.

MLumme:
Uses HttpClient as LajiApiClientB does not support responseType.


rpulkka (approved):
OK!

PR #954, opened 2026-04-08
Named places api v1 migration

olzraiti:
Migrates named place to use the new api client, removing all models and services since the api client takes care of caching.

Adds also "backtracking" for cache flushing to the api client. This fixes flushing parent levels when there isn't a cache entry for the entry that wants to flush. Consider these requests:

```
(1) GET /named-places/
(2) GET /named-places/{id}
(3) POST /named-places/{id}/reservation
```

With the existing cache flushing mechanism, step `(3)` wouldn't invalidate `(1)` and `(2)` because there isn't a cache node for `/named-places/{id}/reservation`. It would work only if there would be a cache entry for `GET /named-places/{id}/reservation`.

The backtracking traverses the cache tree backwards until it finds a cache node, then performs flushing against that.

So at step `(3)` the backtracking does the following:

1. Starts flushing node `/named-places/{id}/reservation`, which doesn't exist
2. Tries `/named-places/{id}` next, which exists
3. Flushes that node and the parent node (`(1)` and `(2)`)


olzraiti:
Also, the `includeUnits` stuff  was removed because it's handled nowadays on API side. Front end doesn't need to check that.

rpulkka (approved):
OK!

PR #955, opened 2026-04-08
Api refactors

Blodir:
couple of notes:
- the `/sample` route doesn't exist in the openapi defs, and it doesn't sound like it's coming soon so forcing it with `any`
- some api routes, like register polygon, have a required person token param. But the api client auto fills that now... I forced any here, but in the future should figure out a proper solution


rpulkka (approved):
OK!

PR #959, opened 2026-04-14
Api client migration: rm person & profile model usage, fixes, refactoring

olzraiti:
* Removes `Person` & `Profile` models
* User service `getProfile()` doesn't return empty profile for empty token. Fail fast instead; the API will throw an error and that is propagated
* Profile component had also defensive coding where it mapped failing API request as empty profiles. Fail fast instead.
* The profile component had a naming scheme for instance variables that was very difficult to follow. The idea is that there are two profiles at play: the logged in user's profile, and the viewed user's profile. I renamed the variables as follows:
	* `this.currentProfile` -> `this.viewedProfile`
	* `this.currentUser` -> `this.loggedInUser`
	* `this.id` -> `this.viewedUserId`
	* `this.userProfile` => `this.loggedInProfile`
* Fixed also some coding errors in the profile component because these two profiles (logged in/viewed) had been mixed apparently.
* I ignored type errors with "any" in document approval because the whole component will be removed by luomus/meta#524


MLumme (commented):
LGTM

PR #964, opened 2026-04-17
API client migration: Line transect and biomon result services

rpulkka:
Some things to consider:

- The warehouse queries seem to return type `string | AggregateResponse` and I tried to handle the string by using `filter((response): response is AggregateResponse => typeof response !== 'string')`
- DESC doesn't seem to be supported in the generated types, so I used `orderBy: ['gathering.conversions.year DESC'] as any`
- I did some type improvements to bird-point-count-result which I pushed last week
- I found out that all but one function were unused in result.service.ts so I rebranded it to result-util.service.ts and deleted the dead code


olzraiti (approved):
LGTM

PR #967, opened 2026-04-23
Api client migration, removing some old models

aorin:
Removes all unused models in /shared/models folder and refactors parts that use Annotation and Audio models to use the new api client models. AudioPlayerComponent had quite much unused code, those were also removed


rpulkka (approved):
Approve!

PR #968, opened 2026-04-24
 API client migration: Red list and informal taxon groups

rpulkka:
The code expects RedListTaxonGroup to have `hasIucnSubGroup?: Array<string|RedListTaxonGroup>;`: https://github.com/luomus/laji/blob/development/projects/laji/src/app/shared/model/RedListTaxonGroup.ts However, the new api.d.ts type RedListTaxonGroupExpanded has `hasIucnSubGroup?: components["schemas"]["IucnRedListTaxonGroupExpanded"];`. This is why I wrote the override to taxon.service.ts.


Blodir (approved):
I'll approve this as the temp solution

PR #978, opened 2026-05-07
laji-taxon-autocomplete pressing enter selects previewed item instead…

olzraiti:
… of the input value

### DIFF
--- a/projects/laji/src/app/shared-modules/taxon-autocomplete/taxon-autocomplete.component.ts
+++ b/projects/laji/src/app/shared-modules/taxon-autocomplete/taxon-autocomplete.component.ts
@@ -61,6 +61,7 @@ export class TaxonAutocompleteComponent implements AfterViewInit, OnDestroy {
 
   private tokenMinLength = 3;
   private destroyBlurListener?: () => void;
+  private previewedTaxon?: any;
 
   constructor(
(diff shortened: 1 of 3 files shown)


projects/laji/src/app/shared-modules/taxon-autocomplete/taxon-autocomplete.component.ts
```diff
@@ -180,11 +181,13 @@ export class TaxonAutocompleteComponent implements AfterViewInit, OnDestroy {
     this.cdr.markForCheck();
   }
 
+  onTaxonPreview(taxon: any) {
```
rpulkka: Would the TypeaheadMatch type be suitable here?:
https://github.com/luomus/laji/blob/development/projects/laji-ui/src/lib/typeahead/typeahead-match.class.ts

rpulkka (changes requested):

projects/laji/src/app/shared-modules/taxon-autocomplete/taxon-autocomplete.component.ts
olzraiti: Yes, added that

rpulkka (approved):
Good!

PR #979, opened 2026-05-07
Fix vihko spreadsheet import for geojson [#472]

### LINKED ISSUE #472: Fix handling polygons with holes
On Vihko/Data Bank, when importing records having polygons with holes, show a validation error ("polygoneissa ei saa olla aukkoja") and prevent saving.

- [ ] Implement
- [ ] Add to Vihko instructions how holes are handled @mikkohei13 
- [ ] Inform Data Bank user: (https://help.helsinki.fi/itsm/EfecteFrameset.do#/workspace/datacard/view/1109776184) @mikkohei13

olzraiti:
Code changes are smaller than it looks at first sight, as I made the control flow easier to read (early guard return instead of massive if block etc).

Basically tries to convert the coordinate column into geoJSON, and if that fails, continues with the existing logic.

Closes luomus/meta#472

### DIFF
--- a/projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
+++ b/projects/laji/src/app/shared-modules/spreadsheet/service/mapping.service.ts
@@ -446,33 +446,39 @@ export class MappingService {
   }
 
   private analyzeGeometry(value: any) {
-    if (typeof value === 'string') {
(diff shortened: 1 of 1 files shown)


rpulkka (approved):
OK!

PR #980, opened 2026-05-08
Replace line breaks in csv/tsv downloads with space [closes #969]

### LINKED ISSUE #969: Lightweight occurrence download and line breaks
Currently making lightweight download here
https://881.dev.laji.fi/observation/list?hasSequenceText=true

as TSV

causes there to be line breaks "content/n
content"

This breaks geoconverter and will cause issues with some users

Lets modify TSV export so that line breaks are replaced by " "  (space)

aorin:
https://github.com/luomus/laji/issues/969
https://969.dev.laji.fi/observation/list?hasSequenceText=true

This is based on branch `dna-seq-filter-881` so that it's easier to test


MLumme (approved):
LGTM

PR #984, opened 2026-05-26
Migrate graphql to api v1

olzraiti:
https://260522.dev.laji.fi


olzraiti:
works locally but not on the feature branch. Most of the time working on this has went to resolving network issues, sigh

olzraiti:
Ok it wasn't a network issue. Something goes silently wrong when a header is undefined - fixed that. Person token header was undefined when logged out

Blodir (approved):

PR #1010, opened 2026-07-09
Archipelago bird census result service [Closes #436]

rpulkka:
Issues: https://github.com/luomus/meta/issues/436 https://github.com/luomus/meta/issues/442
Branch: https://436.dev.laji.fi/project/MHL.1166/stats?tab=statistics

Implements the archipelago bird census result service. It has two tabs: /stats for the specific result service and /map for the generic result service. Map uses the existing laji-result-map component. Stats is a new component, which lets the user select year, route and np and then a table of user's own observations is shown. As stats show user's own observations, login is required, unlike in map.

getRowItems$ pipe gets the taxonomic order, appends dateBegin and gatheringType to rows, includes only observationStatusObserved, constructs a observation string which includes shorthands of all possible counts, orders primarily by taxonomic order and secondarily by dateBegin, adds gatheringType "icon" as unicode, and keeps taxon name only on the first unique row of that name.


MLumme (approved):
LGTM

Issue #1017, opened 2026-08-05
Fix handling of expired person token


rpulkka:
On dev / beta, when person token has been expired, instead of logging out the user can visit pages for logged in users and gets a lot of 400 errors. Also "External - Outgoing request failed without message".

<img width="1917" height="945" alt="Image" src="https://github.com/user-attachments/assets/78845ecc-88c9-442a-a765-92cbd3bb23a2" />

PR #1020, opened 2026-08-10
confirm persontoken validity before setting api client to use it

Blodir:
When the apiclient was updated to auto-fill the persontoken, the token was stored in apiclient before the validity of the token was confirmed as the validity check itself needed the persontoken. Fixed by allowing the apiclient user to overwrite the token header. Now the token is used only for the validity check, and then stored in the apiclient only after the check passes.


Blodir:
Fixes https://github.com/luomus/laji/issues/1017

olzraiti (approved):

PR #1028, opened 2026-08-17
Remove internal observation search form state from warehouse queries

Blodir:
^title

unfortunately the observation form has been structured in such a way that the internal query object doesn't actually match what API accepts. This PR runs all the warehouse queries used in observation search through normalization


Blodir:
Fixes https://github.com/luomus/laji/issues/1029

rpulkka (approved):
OK!

Issue #1030, opened 2026-08-18
Stop taxon overview from making data warehouse queries


Blodir:
> I think currently taxon card overview tab selects image so that
> 1. use taxon image
> 2. use occurrence image
> 3. use specimen image
> 
> Bug:  The overview selection logic has not been updated when we removed COMMUNITY VERIFIED images from the > Media tab, it is still using COMMUNITY VERIFIED images
> Problem: We have disabled showing occurrence and specimen images for non-logged in users for performance > reasons. BUT we didn't remember to disable overview tab from making a /warehouse query when resolving the primary image

https://luomus-ict.slack.com/archives/C014PNS2S3F/p1786968828754739?thread_ts=1786954306.928209&cid=C014PNS2S3F

Blodir:
Taxon overview now uses the same fallback queries as media tab, but only if you are logged in

testable: https://5934085.dev.laji.fi/

test instructions:
- new tab -> f12 -> open network tab
- login https://5934085.dev.laji.fi
- go to https://5934085.dev.laji.fi/taxon/MX.277915?showTree=true
- filter by `unitMedia` to see the queries
- switch to media tab to see that there are no duplicates

PR #1031, opened 2026-08-18
Unify taxon overview media queries

Blodir:
https://github.com/luomus/laji/issues/1030

https://luomus-ict.slack.com/archives/C014PNS2S3F/p1787055100337229?thread_ts=1786954306.928209&cid=C014PNS2S3F

Note: previously committed the associated login check directly to dev/beta

This PR updates the taxon overview image fallback queries to use the same filters as media page.

Also unified the page size to 12 so that both queries hit the same cache item.


Blodir:
It shows a couple of extra commits, but that's a github bug caused by this branch being based off of beta (and then merging beta and dev later). Only the last 4 are relevant for this PR

Edit: I got them to disappear by changing the base branch back and forth

olzraiti (approved):

PR #1032, opened 2026-08-20
Ssr build can be ran locally with start:local-build

olzraiti:
Adds new script `start:local-build` that starts the local build (built with `npm run build:local`)

Removes some broken npm scripts also.


Blodir:
Nice! I made the script use cross-env for windows compat, but otherwise looks good

PR #1049, opened 2026-09-08
Fix IUCN userService state

Blodir:
Base graphql query stalled because userService.loggedIn$ didn't emit a value. Fixed by initializing userService as notLoggedIn when env type is iucn


Blodir:
Fixes https://github.com/luomus/laji/issues/1048

olzraiti (approved):

PR #1050, opened 2026-09-09
Refactor data bank page to make only one form permission request

olzraiti:
Atm it does one request per databank. There's an API endpoint which returns all collections that a person has permission to, so we use it instead here.


projects/laji/src/app/shared/service/form-permission.service.ts
```diff
@@ -141,29 +142,24 @@ export class FormPermissionService {
...
-          switchMap(person => this.getFormPermission(collectionID).pipe(
-            catchError(() => of({
-              collectionID,
-              admins: [],
-              editors: []
-            } as unknown as FormPermission)),
-            map((formPermission: FormPermission) => ({person, formPermission}))
-          )),
-          switchMap(({person, formPermission}) => person ? of({
-            view: this.isEditAllowed(formPermission, person, form) || form.options?.restrictAccess === 'MHL.restrictAccessLoose',
```
aorin, line 159: `isEditAllowed` could be removed as the function is not used anymore. Otherwise looks good

aorin (approved):

PR #1054, opened 2026-09-10
Typeahead improvements

### LINKED ISSUE #899: Fix taxon dropdown positioning on mobile
On observation search at https://laji.fi/observation
when using mobile device, the taxon dropdown takes too much space horizontally, so it's mostly blank, unless user scrolls horizontally.

Used to work better, I don't know when this has changed. Appeared in Jan 2026 or before.

Tested: 
- Firefox on Android (image)
- Safari on iOS

Image

Blodir:
Fixes #899 

- Removed a whole bunch of dead code from lu-typeahead (cruft that came with when it was copypasted from ngx-bootstrap)
- also changed the placement option to use `PlacementService` directly
- Fixed the aforementioned issue by allowing the dropdown to resize


projects/laji-ui/src/lib/placement/placement.service.ts
olzraiti, line 86: Did some experimenting with this. `throttleTime(0)` seems to be good enough in practice. It accepts a scheduler though, so in theory `throttleTime(0, animationFrameScheduler)` could be more optimal... anyway, in practice this is good enough

olzraiti (approved):

PR #1057, opened 2026-09-14
Fix own occurence search multiplication 870

### LINKED ISSUE #870: Fix own occurrence search query duplication/triplication
This bug has performance implications. It causes frontend to do the exact same query 2 or 3 times.

Image

If you select first "Havainnoija" queries are executed only once
If you then select also "Muokkaajana" queries are executed twice
if you then unselect "Havainnoijana" queries are executed three times

MLumme:
Fixes #870 

Testable here https://870.dev.laji.fi/observation/list, fixes both the request multiplication and user not being able to remove the edotorOrObserverToken search parameter using the own observations element if both were selected and then both removed.

Note that this only concerns aggregate- and list-requests, and there are naturally two very similar aggregate-requests, but there are other duplicates now for some reason.


rpulkka (approved):
OK!

PR #1058, opened 2026-09-14
Project form fail fast

### LINKED ISSUE #736: Fix form not being displayed for editor
Repro:

- With user that has admin rights to a data bank collection:
  - Add a document to a primary data bank collection, e.g. https://dev.laji.fi/project/MHL.617/about
  - Add your friend (who is neither ict-admin, nor admin of that dataset) as observer, and give them editor rights
  - Save the document
- With the friend user:
  - Find that document
  - Click Edit button

Expected:

- User can edit the document, since they have editor rights

Actual:

- Form is empty

olzraiti:
Fixes #736

I simplified the code so that it doesn't try to model different error situations and provide a message deduced on front-end code. Instead we fail fast and let API errors bubble to the error handler that displays the error.

API already allowed accessing document that the person is editor of, so letting API handle permissions is sufficient.

### DIFF
--- a/projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
+++ b/projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
@@ -1,5 +1,5 @@
 import { Injectable } from '@angular/core';
-import { catchError, map, mergeMap, switchMap, take, tap, filter, distinctUntilChanged, mapTo, shareReplay, distinctUntilKeyChanged } from 'rxjs';
+import { catchError, map, mergeMap, switchMap, take, tap, distinctUntilChanged, mapTo, shareReplay, distinctUntilKeyChanged } from 'rxjs';
 import { combineLatest, concat, merge, Observable, of, ReplaySubject, Subscription } from 'rxjs';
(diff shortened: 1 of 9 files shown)


projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
```diff
@@ -107,109 +86,82 @@ export class DocumentFormFacade {
...
-                }
-                return {form, namedPlace, formData: documentModel.document, hasChanges: documentModel.hasChanges, template};
-              }))
+              ).pipe(map(documentModel =>
+                ({form, namedPlace, formData: documentModel.document, hasChanges: documentModel.hasChanges, template})))
             )
           )
-      )
+      ),
+      shareReplay()
```
aorin, line 137: I think this should be `shareReplay(1)` (that stores and emits only the last value to new subscribers). Without the buffer size it stores all values and emits them all to new subscribers. I guess it doesn't matter much in this case but usually it's good to have the buffer size if there is no specific need to store all the values

aorin (approved):
The code is clearer now 👍

PR #1063, opened 2026-09-14
Fix annotation view translation

### LINKED ISSUE #898: Fix missing translation on annotation view
### Steps to reproduce

1. Go to see annotation view of unit of a document that has multiple units
2. See text "15 annotation.observations.count"

Image

### Expected Behaviour

Should localize "annotation.observations.count" on all languages

### Additional information

_No response_

aorin:
Fixes #898

https://898.dev.laji.fi/observation/list?annotated=true

The translation key `annotation.observations.count` doesn't exist and the `annotation.observation.count` is in a format that works for both one and multiple observations ("7 observation(s) in this document") so I just removed the `annotation.observations.count` from the code


rpulkka (approved):
OK!

PR #1064, opened 2026-09-14
Fix label pipe with an invalid url

### LINKED ISSUE #1026: Document viewer fails to show facts correctly if it can't understand a URI-address as fact value
### Steps to reproduce

1. Go to 'https://beta.laji.fi/view?uri=http:%2F%2Ftun.fi%2FJX.349699'
2. The document has unit.facts: fact: `http://tun.fi/MY.sequenceText value: http://tun.fi/MA.5XXXX`
3. UI shows it as blank

### Expected Behaviour

If trying to display URI into some content fails, the UI should show the value as it is

### Additional information

Image

aorin:
Fixes #1026

https://1026.dev.laji.fi/view?uri=http:%2F%2Ftun.fi%2FJX.349699

Fixes label pipe returning an empty string if the url is invalid

### DIFF
--- a/projects/laji/src/app/shared/pipe/label.pipe.ts
+++ b/projects/laji/src/app/shared/pipe/label.pipe.ts
@@ -85,7 +85,9 @@ export class LabelPipe implements PipeTransform, OnDestroy {
         );
       case 'fullUri':
         return key.indexOf('http') === 0 ?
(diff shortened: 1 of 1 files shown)


rpulkka (approved):
Good!

PR #1069, opened 2026-09-15
Fix autocomplete style issues [Fixes #776, Fixes #987]

### LINKED ISSUE #776: Fix taxon autocomplete bold text styling
**Bug 1:**
There is an extra white space after the ( -char
Image

**Feature 2:**
(Btw, this is not a bug - this is spec change; can be implemented if easy or disregarded)

Go to https://beta.laji.fi
Type "tali" to header taxon search
The active selection has everything bolded (div class=active: font-weight:700)  so the bolding it is not showing up.
 

Image

Fix is to do the same than on other taxon autocomplete: use background color instead of bolding to show the selected row. There already is a background color, so this font-weight: 700 could just be removed from .taxon-matches div.active

### LINKED ISSUE #987: Taxon autocomplete doesn't work for colloquialVernacularName matches
### Steps to reproduce

1. Go to https://laji.fi/observation
2. Write "mustaraitanaamiokärpänen" to Lajin nimi
3. Click the name
4. Expected: name appears on the filter
5. Actual: name doesn't appear

This is caused by the UI showing up somehow the name in the language that was typed

Type: Aves -> Aves shows up, target filter is MX.37580
Type: nisäkkäät -> nisäkkäät shows up (and not Mammalia), target filter is MX.37612

The first match for "mustaraitanaamiokärpänen" is  colloquialVernacularName of  MX.278297 that doesn't have a vernacular name -> ui doesn't know what to show
In these cases, the UI could always show the scientific name and if not present (odd cases), the MX.xxx (or gbif:xx) identifier

### Expected Behaviour

_No response_

### Additional information

_No response_

rpulkka:
Issue 1: https://github.com/luomus/laji/issues/776
Issue 2: https://github.com/luomus/laji/issues/987
Branch: https://776.dev.laji.fi/observation

Autocomplete style fixes:

- No extra space after "("
- Result bolds only the matching parts, not the whole result
- Fix fallback logic from `taxon['vernacularName'] !== ''` to `taxon['vernacularName'] && taxon['vernacularName'] !== ''` to show scientificName for undefined vernacularNames

### DIFF
--- a/projects/laji/src/app/shared/service/taxon-autocomplete.service.ts
+++ b/projects/laji/src/app/shared/service/taxon-autocomplete.service.ts
@@ -114,12 +114,15 @@ export class TaxonAutocompleteService {
       case 'MX.euringCode':
         return taxon['matchingName'].toLowerCase();
       case 'MX.vernacularName':
-        return taxon['vernacularName'] !== '' ? taxon['vernacularName'] : taxon['scientificName'];
       case 'MX.alternativeVernacularName':
       case 'MX.obsoleteVernacularName':
       case 'MX.tradeName':
       case 'MX.colloquialVernacularName':
-        return taxon['vernacularName'] !== '' ? taxon['vernacularName'] : taxon['scientificName'];
+        return (taxon['vernacularName'] && taxon['vernacularName'] !== '')
+          ? taxon['vernacularName']
+          : (taxon['scientificName'] && taxon['scientificName'] !== '')
+            ? taxon['scientificName']
+            : taxon['id'];
       default:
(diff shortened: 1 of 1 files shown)


olzraiti (changes requested):
In all the `taxon['vernacularName'] && taxon['vernacularName'] !== ''` checks the `taxon['vernacularName'] !== ''` is redundant. If `taxon['vernacularName']` is truthy, it can't be `''`.

So could be simplified to just check for `taxon['vernacularName']`

olzraiti (approved):
thanks!

PR #1073, opened 2026-09-15
Add tmp id to template documents [Fixes #562]

### LINKED ISSUE #562: Mobilise Syke moth census data
Valtakunnallinen yöperhosseuranta (Nocturna), which has data stored in YÖPETI-database.

- As Nocturna project admin, I want to collect and maintain data using Vihko, so that data is managed better than now.

- As volunteer entering data, I want to have a good data entry UI, to avoid errors and save time.

- As Nocturna project admin, I want that systematic errors in the data are fixed, and further errors are avoided, so that the data will have good quality.

- As Nocturna project admin, I want my data to be available on http://laji.fi/ , so that it can be used by others.

rpulkka:
Issue: https://github.com/luomus/meta/issues/562
Branch: https://562.dev.laji.fi/vihko/home

Because documents that are created from a template are not considered new, they ended up in `fetchExistingDocuments` and were never assigned a temporary ID. Therefore if user quit the browser, there was no record on unsaved events. Now `fetchExistingDocuments` adds a temporary ID to documents without an ID.

### DIFF
--- a/projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
+++ b/projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
@@ -377,7 +377,10 @@ export class DocumentFormFacade {
         mergeMap(local => this.documentService.findById(documentID).pipe(
           map((document: Document) => {
             if (document.isTemplate) {
-              const doc = this.documentService.removeMeta(document, ['isTemplate', 'templateName', 'templateDescription']);
+              let doc = this.documentService.removeMeta(document, ['isTemplate', 'templateName', 'templateDescription']);
+              if (!doc.id) {
+                doc = { ...doc, id: this.getNewTmpId() };
+              }
               return {
                 document: form.options?.prepopulatedDocument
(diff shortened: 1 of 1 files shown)


projects/laji/src/app/project-form/form/document-form/document-form.facade.ts
```diff
@@ -377,7 +377,10 @@ export class DocumentFormFacade {
         mergeMap(local => this.documentService.findById(documentID).pipe(
           map((document: Document) => {
             if (document.isTemplate) {
-              const doc = this.documentService.removeMeta(document, ['isTemplate', 'templateName', 'templateDescription']);
+              let doc = this.documentService.removeMeta(document, ['isTemplate', 'templateName', 'templateDescription']);
+              if (!doc.id) {
```
aorin: The previous line always removes the id so the check `if (!doc.id) {` could be removed

aorin (changes requested):

aorin (approved):
👍

